PR #29, opened 2018-06-26
Bug fix: push correct has-pr tag

AdamHillier:
I could be wrong but I think this got missed in 578d63ae669b3e30336a6c4b12f3041dd82cd4c2?


k88hudson:
Thanks!

PR #32, opened 2018-06-26
Fix #28: Add 'chore' tags

### LINKED ISSUE #28: Add support for 'chore' tags
To indicate a chore, we are decorating bugzilla tickets with the `whiteboard` field set to `[chore]`

Let's change bugzy so this  is displayed as a bugzy tag.

AdamHillier:
Wasn't sure about the style of the chore tags - these are purple on white

### DIFF
--- a/src/config/tags.js
+++ b/src/config/tags.js
@@ -5,7 +5,7 @@ module.exports = {
     style: {color: "var(--red-50)", fontWeight: "bold"},
     label: "defect"
   },
-  "NEEDINFO": {
+  "needinfo": {
     style: {color: "var(--gray-90)", fontWeight: "bold"},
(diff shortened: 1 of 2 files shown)


src/config/tags.js
```diff
@@ -32,5 +32,9 @@ module.exports = {
   "strings m-c needed": {
     label: "strings m-c needed",
     style: {backgroundColor: "var(--teal-80", color: "white", border: 0, fontWeight: "bold"}
+  },
+  "chore": {
+    label: "chore",
+    style: {color: "var(--purple-60)", fontWeight: "bold"}
```
k88hudson: let's flip it around so purple is the background

PR #34, opened 2018-06-26
Fix #15: Add a way to flush metabug cache

### LINKED ISSUE #15: Add a way to force-refresh cached meta bugs
Right now they are cached for 3 hours with no way to refresh them

AdamHillier:
There's no icon at the moment, but this was the most straightforward solution I could think of.

### DIFF
--- a/src/content/components/Router/Router.js
+++ b/src/content/components/Router/Router.js
@@ -27,6 +27,12 @@ const RouterNav = withRouter(class _RouterNav extends React.PureComponent {
       {route.label}
     </NavLink></li>);
   }
+
+  async refreshMetas() {
+    await fetch('/refresh_metas');
+    window.location.reload();
+  }
+
   render() {
     const {routes} = this.props;
     return (<nav className={styles.aside}>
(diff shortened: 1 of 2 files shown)


src/server/server.js
```diff
@@ -39,6 +39,12 @@ app.get('/api/metas', async (req, res) => {
   res.send(metasCache.data);
 });
 
+app.get('/refresh_metas', (req, res) => {
+  metasCache.data = null;
+  metasCache.lastUpdated = null;
+  res.redirect('/');
```
k88hudson: this is OK but results in you navigating away from your current page, how about just `res.end()` (which should send a `200` response) and then you can `fetch`/force refresh the current page?

AdamHillier:
So this seems to work as suggested though I'm a little unsure how/if/what error handling should be done with the fetch call.

src/content/components/Router/Router.js
```diff
@@ -27,6 +27,11 @@ const RouterNav = withRouter(class _RouterNav extends React.PureComponent {
       {route.label}
     </NavLink></li>);
   }
+
+  refreshMetas() {
+    fetch('/refresh_metas').then(location.reload);
```
k88hudson: I would do this:

```js
async refreshMetas() {
  await fetch("/refresh_metas");
  window.location.reload();
}

just to prevent the `response` from fetch being passed to `reload()`. (`reload(true)` actually causes the page to be force-reloaded, which is not really necessary here)
```

AdamHillier: Yes that's a good point, thank you!

PR #36, opened 2018-06-27
Fix #12: add button to copy metabug number

### LINKED ISSUE #12: make feature meta bug number easily copy-able
remembering meta bugs is annoying, so having this would be nice

AdamHillier:
image

It looks like this. Not 100% about the styling but it seems to be functional at least.

I didn't realise that copying text from JS was so complicated, but I've discovered that there's no nice API to use; the thing I've included here was the cleanest and least nasty way that I could find to do it. (I'm also not 100% sure about correct code attribution.)

Also, the icon is from [here](https://www.flaticon.com/free-icon/copy-file_73547#term=copy&page=1&position=22) which is an existing source of some of the other icons.


k88hudson:
This works great, I think it needs some visual feedback though; maybe just show a small absolute positioned bit of text for a few seconds after the copy happens?

PR #37, opened 2018-06-28
Fix #8: Add eslint

### LINKED ISSUE #8: Add eslint

AdamHillier:
Sorry, this is quite a large patch.... Broadly speaking it's the eslint rules from ActivityStream.

A few things:
- I'm not linting the two files with flow types in them at the moment, but I can add that if you'd prefer.
- I've made "alphabetically sort imports" a warning as there's no automatic way to sort them and I feel like that rule is a little strict
- I've made "react/no-deprecated" a warning as I didn't want to get too involved with code changes - maybe we can file another ticket for that?
- I've used Husky for a pre-push git lint task.
- There were quite a few errors about things being defined and not used; in some of these cases I've just commented out the thing that was defined in case you want to keep access to it (e.g., `// const EPIC_BUG_NUMBER = ...`).
- Almost all the fixes were from `eslint --fix`.


Mardak:
> Almost all the fixes were from eslint --fix.

Perhaps in the future, for easier review, you could put all the autofixed stuff in its own commit. Typically I do this by adding the .eslintrc changes and --fix then adding the changes except eslintrc changes.

package.json
```diff
@@ -12,7 +12,9 @@
     "prebundle": "npm run remove_types",
     "bundle": "parcel build src/content/index.html --out-dir dist/content --public-url \"/\"",
     "deploy": "git checkout -B deploy && npm run bundle && git add -f dist",
-    "test": "npm run flow && mocha --require babel-register src/**/*.test.js"
+    "test": "npm run flow && mocha --require babel-register src/**/*.test.js",
```
k88hudson: actually can you run lint in test as well?

k88hudson:
(feel free to merge after fixing that nit)

PR #59, opened 2018-08-20
Fix Bug 58 - Add Exception for 63 long cycle

piatra:
Blocker: Couldn't see what bugs we have planned 😅


k88hudson:
@piatra thanks!

PR #92, opened 2019-02-01
Set needuplift status when cf_tracking_firefox66 is question mark

punamdahiya:


punamdahiya:
needuplift status should show up right after triage when a bug is marked 66 affected with cf_tracking_firefox<release> as '?' e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1524321

PR #95, opened 2019-02-20
P1/P2 definition changes andr related query changes

k88hudson:
Also adds uplift view for regular feature page.

### DIFF
--- a/src/content/components/FeatureView/FeatureView.js
+++ b/src/content/components/FeatureView/FeatureView.js
@@ -11,6 +11,7 @@ import {BUGZILLA_PRODUCT, FILE_NEW_BUGZILLA_COMPONENT} from "../../../config/pro
 const currentIteration = getIteration().number;
 const currentRelease = currentIteration.split(".")[0];
 const prevRelease = parseInt(currentRelease, 10) - 1;
+const nextRelease = parseInt(currentRelease, 10) + 1;
 
 const upliftTrackingField = `cf_tracking_firefox${prevRelease}`;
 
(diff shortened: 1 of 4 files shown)


src/content/components/PocketNewtabView/PocketNewtabView.js
```diff
@@ -207,7 +207,7 @@ export class PocketNewtabView extends React.PureComponent {
     for (const bug of bugs) {
       if (bug.summary.toLowerCase().startsWith("[meta]")) {
         continue;
-      } else if (["P1", "P2"].includes(bug.priority)) {
+      } else if (bug.priority === "P1") {
```
punamdahiya: change here is pushing P2 bugs fixed earlier in backlog which doesn't look right

punamdahiya: For Pocket New Tab we should use both P1 and P2 as it shows consolidated view of multiple releases

src/content/components/FeatureView/FeatureView.js
```diff
@@ -121,12 +127,19 @@ export class FeatureView extends React.PureComponent {
...
+
       <h3>Uplift candidates</h3>
       <BugList showResolvedOption={false} bulkEdit={true} tags={true} bugs={bugsByRelease.uplift} columns={[...displayColumns, "cf_status_nightly", "cf_status_beta"]} />
 
       <h3>Required for Current Release (Firefox {currentRelease})</h3>
       <BugList showResolvedOption={false} bulkEdit={true} tags={true} bugs={bugsByRelease.current} columns={displayColumns} />
 
+      <h3>Required for Next Release (Firefox {nextRelease})</h3>
+      <BugList showResolvedOption={false} bulkEdit={true} tags={true} bugs={bugsByRelease.next} columns={displayColumns} />
+
```
punamdahiya: WIth each priority its owen section, do we need priority column (Lin 25) in feature view?

k88hudson: ah good point

k88hudson: for backlog it's useful since it can have p5 as well

src/content/components/PocketNewtabView/PocketNewtabView.js
k88hudson: ok I'll leave it as is for now, let's discuss tomorrow

src/content/components/FeatureView/FeatureView.js
k88hudson: Yeah I think we should just leave it for now while people are getting used to the p1/p2/p3 changes also

PR #98, opened 2019-02-22
Add tabs UI to feature view and other adjustments

k88hudson:
Adds some basic tab UI that splits up FeatureView into 2 sections for
now; this is pretty rough but we can expand in the future (for example,
it would be nice to split up the queries to actually only fetch resolved
bugs if we have to.)

Also changes the BugLists to use the "compact" styles and sorts resolved
bugs by resolved date (cf_last_resolved).


k88hudson:
@punamdahiya do you mind taking a look at this when you get a chance?

PR #104, opened 2019-03-12
Add Container and Tabs component

k88hudson:
This patch adds a `Container` component that we can use to wrap pages as well as a `Tabs` component that could be reused for adding a tab-based UI to other pages.


punamdahiya:
r+, thanks for the refactoring. LGTM!

PR #108, opened 2019-03-22
Fix Bug 107 - Bulk edit all bugs on page

piatra:


src/content/components/BugList/BugList.js
```diff
@@ -81,9 +81,17 @@ export class BugList extends React.PureComponent {
...
   renderBulkEdit(selectedBugs) {
-    return (<EditorGroup>
-      <a className={gStyles.primaryButton} href={this.getBulkEditLink(selectedBugs)}>Edit in Bugzilla</a>
-    </EditorGroup>);
+    const allBugsOnPage = [...document.querySelectorAll("input[value]")]
+      .map(i => i.value)
+      .filter(v => parseInt(v, 10));
+    return (<React.Fragment>
+      <EditorGroup>
+        <a className={gStyles.primaryButton} href={this.getBulkEditLink(allBugsOnPage)}>Bulk edit all</a>
```
punamdahiya: In future we can get design take on bringing 'bulk edit all' at the top instead of one per section.

PR #110, opened 2019-04-02
Sort bugs in FeatureView by subMeta

k88hudson:
Adds some separation of bugs by meta in the engineering view, only for open P1 bugs for now.

Try this by taking a look at http://localhost:1989/feature/1536844 for example, notice how sub metas now have their own sections

### DIFF
--- a/src/content/components/FeatureView/FeatureView.js
+++ b/src/content/components/FeatureView/FeatureView.js
@@ -33,6 +33,7 @@ const allColumns = displayColumns.concat([
   "keywords",
   "severity",
   "flags",
+  "blocks",
   upliftTrackingField,
   `cf_status_firefox${prevRelease}`,
   `cf_status_firefox${currentRelease}`,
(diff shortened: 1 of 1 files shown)


src/content/components/FeatureView/FeatureView.js
```diff
@@ -71,14 +72,19 @@ const FeatureBugList = ({hideIfEmpty, bugs, title, extraColumns = [], ...restPro
...
   return (<React.Fragment>
     <FeatureBugList title="Untriaged bugs" hideIfEmpty={true} bugs={bugs.untriaged} extraColumns={["cf_status_nightly", "cf_status_beta"]} />
-
     <FeatureBugList title="Uplift candidates" hideIfEmpty={true} bugs={bugs.uplift} extraColumns={["cf_status_nightly", "cf_status_beta"]} />
-    <FeatureBugList title={`Required for Current Release (Firefox ${currentRelease})`} bugs={bugs.current} />
+    <h3>Required for Current Release (Firefox {currentRelease})</h3>
+    {subMetas.length ? [
+      ...Object.keys(bugs.currentBySubMeta).map(id => {
+        const meta = subMetas.find(m => String(m.id) === id);
+        return <FeatureBugList key={meta.id} subtitle={meta.summary} showHeaderIfEmpty={true} bugs={bugs.currentBySubMeta[meta.id]} />;
```
punamdahiya: We can think of making subtitle here a link so that its easier to navigate to sub meta bug in bugzilla. Right now one need to open one of the bugs to find sub meta bug id

PR #113, opened 2019-04-24
Set new export bug type to task

Mardak:
r?@k88hudson I've been creating export as task to make the depends list show it separately from enhancement bugs. This won't work correctly until https://bugzilla.mozilla.org/show_bug.cgi?id=1543456 is deployed to bmo.


Mardak:
Sounds like there will be a per-component default type setting if we want to rely on that instead of changing our other bug creation urls. Probably enhancement instead of defect, so we'll still need this PR to explicitly set as task. https://bugzilla.mozilla.org/show_bug.cgi?id=1546788

PR #114, opened 2019-04-25
Bug 112 - Use the bugzilla 'type' field to label bugs in the correct category

piatra:
image
image
image
There's some overlap with how we previously used to mark `chore`s (using the whiteboard tag) so this also shows up:
image
But that will go away as we stop using it.


src/schema/query_options.js
```diff
@@ -193,7 +194,7 @@ const custom_filters = [
   "dependson",
   "whiteboard",
   "creation_ts",
-  "severity",
+  "type",
```
punamdahiya: I don't think we are sorting by type, with severity removed we should remove "bug_severity" from sort_keys below

punamdahiya:
https://github.com/k88hudson/bugzy/blob/master/src/schema/customfields.json#L323 is redundant now, we should remove and add type.

punamdahiya:
Landing PR with nits to be covered in a follow up

PR #132, opened 2019-05-30
Bug 57 - Make priority sidebar collapsable

AllegroFox:
It toggles! Still have to work on replacing the button with a nice icon.

### DIFF
--- a/src/content/components/PriorityGuide/PriorityGuide.scss
+++ b/src/content/components/PriorityGuide/PriorityGuide.scss
@@ -37,8 +37,9 @@
 .sidebarToggle {
   position: absolute;
   right: 10px;
-  border-radius: 5px;
-  height: 15px;
+  border: 0;
+  background: none;
+  height: 20px;
 }
 
 .active {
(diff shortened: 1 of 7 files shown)


src/content/components/PriorityGuide/PriorityGuide.scss
```diff
@@ -1,6 +1,5 @@
 .container {
   padding: 20px;
-  background: #f9f9fa;
   width: 250px;
```
k88hudson: You're also going to want to transition the width so this shrinks along with the inner content

src/content/components/PriorityGuide/PriorityGuide.scss
```diff
@@ -34,3 +33,57 @@
...
+.active {
+  background: #f9f9fa;
+  padding: 20px;
+  position: absolute;
+  width: 250px;
+  top: 0px;
+  right: 0;
+  bottom: 0;
+
+  -moz-transition: right 0.25s ease;
```
k88hudson: no need for `moz-transition`! I hope no one is using <Firefox 16

k88hudson: Also,  using `translateX` here is better for performance than using `right` (see this article for more if you're interested https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/)

AllegroFox: Is that something you'd like me to change? Or just an FYI :)

k88hudson: Please change yeah

AllegroFox: okay, will do!

PR #141, opened 2019-06-12
Bug 129 - Make "submetas" work with all sections, not just `p1`

AllegroFox:
I've disabled the no-lonely-if eslint rule on line 379 for now - to me it's clearer if each branch has the same structure in this case, but I can always refactor it!

### DIFF
--- a/src/content/components/FeatureView/FeatureView.js
+++ b/src/content/components/FeatureView/FeatureView.js
@@ -365,7 +365,9 @@ export class FeatureView extends React.PureComponent {
         if (subMetas.length) {
           let subMetaMatch = subMetas.filter(m => bug.blocks.includes(m.id));
           if (subMetaMatch.length) {
-            subMetaMatch.forEach(m => result.untriagedBySubMeta[m.id].push(bug));
+            subMetaMatch.forEach(m =>
+              result.untriagedBySubMeta[m.id].push(bug)
+            );
           } else {
             result.untriaged.push(bug);
           }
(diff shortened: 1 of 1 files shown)


AllegroFox:
I don't know why the Travis build is failing here by the way...it's a Prettier failure, and everything is passing locally.

Mardak:
Did you `npm i` ? I see the same error:

```
Checking formatting...
src/content/components/FeatureView/FeatureView.js
Code style issues found in the above file(s). Forgot to run Prettier?
```

Mardak:
Ah, looks like the pattern is to `npm run prettier` and it generates this diff:
```diff
diff --git a/src/content/components/FeatureView/FeatureView.js b/src/content/components/FeatureView/FeatureView.js
--- a/src/content/components/FeatureView/FeatureView.js
+++ b/src/content/components/FeatureView/FeatureView.js
@@ -367,3 +367,5 @@ export class FeatureView extends React.PureComponent {
           if (subMetaMatch.length) {
-            subMetaMatch.forEach(m => result.untriagedBySubMeta[m.id].push(bug));
+            subMetaMatch.forEach(m =>
+              result.untriagedBySubMeta[m.id].push(bug)
+            );
           } else {
```

src/content/components/FeatureView/FeatureView.js
```diff
@@ -248,10 +286,13 @@ export class FeatureView extends React.PureComponent {
   sortByRelease(bugs, subMetas) {
     const result = {
       untriaged: [],
+      untriagedBySubMeta: {},
```
k88hudson: I think actually it's ok if `untriaged` doesn't get split up by sub meta, there's usually not enough bugs to warrant spliting them up

PR #143, opened 2019-06-19
Add Messaging System to triage components

k88hudson:


Mardak:
Do we want to add New Tab Page now or I suppose we can do that when we remove the activity streams one.

k88hudson:
Let's wait (since we'll need to deal with a bunch of new untriaged stuff)

PR #144, opened 2019-06-25
Bugzilla migration prep

AllegroFox:
This does most of the work for prepping Bugzy to correctly sort and display bugs based on the new components we will have after migration. New tabs, adjusted queries, and a modified sorting function (plus a padding fix for the priority sidebar.)


k88hudson:
Already merged!

PR #146, opened 2019-06-26
bug 145 - Update file new bug buttons

AllegroFox:
Modified the FileNewBugButton to check the bug meta for the correct component; minor query refactor

### DIFF
--- a/src/config/project_settings.js
+++ b/src/config/project_settings.js
@@ -12,7 +12,7 @@ module.exports = {
   // Old epic bug blocking <2019 projects/features
   LEGACY_EPIC_BUG_NUMBER: "1433315",
   // What component should new bugs be filed in?
-  FILE_NEW_BUGZILLA_COMPONENT: "Activity Streams: Newtab",
+  // FILE_NEW_BUGZILLA_COMPONENT: "Activity Streams: Newtab",
   // Metabug for triaging pockt bugs
   POCKET_META: 1535711,
   // Used for some descriptions / titles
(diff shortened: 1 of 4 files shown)


src/config/project_settings.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   // Old epic bug blocking <2019 projects/features
   LEGACY_EPIC_BUG_NUMBER: "1433315",
   // What component should new bugs be filed in?
-  FILE_NEW_BUGZILLA_COMPONENT: "Activity Streams: Newtab",
+  // FILE_NEW_BUGZILLA_COMPONENT: "Activity Streams: Newtab",
```
k88hudson: Can you delete this instead of commenting it out?

AllegroFox: yep!

src/content/components/FeatureView/FeatureView.js
```diff
@@ -124,7 +124,9 @@ const EngineeringView = props => {
                 key={meta.id}
                 subtitle={removeMeta(meta.summary)}
                 meta={meta.id}
-                fileNew={`blocked=${meta.id}, ${parentMeta}`}
+                fileNew={`blocked=${meta.id},${parentMeta}&component=${
+                  parentMeta.component
```
k88hudson: can you change this to `meta.component` since `parentMeta` is apparently a bug number

src/config/project_settings.js
Mardak: Looks like this was deleted in https://github.com/k88hudson/bugzy/commit/33f726b1e68d0e13d104a41d1bd8fa778d3b65c6 but got restored from a force push and was committed:

https://github.com/k88hudson/bugzy/commit/4fa4b08aac093e453f92c4e4a466ac1b6399be7f#diff-75d6aa8e05b9123d4745966a7271dfd4L15

PR #179, opened 2021-05-17
Fix long Nightly dates for 89+, add 92

AllegroFox:


k88hudson (approved):

k88hudson (changes requested):
should be 90

src/bin/buildRange.ts
```diff
@@ -66,6 +66,14 @@ const REFERENCE_RELEASES: ReleaseDefinition[] = [
     startVersion: 87,
     iterationsPattern: [2, 2],
   },
+  {
+    startVersion: 89,
```
k88hudson, line 70: This is for nightly, so this should be 90 right?

AllegroFox: oh, yep.

k88hudson (approved):

PR #189, opened 2022-09-30
Introduce caching for longer loading views

### LINKED ISSUE #188: Introduce caching for longer loading views
Caching data for views such as current iteration and triage would be helpful when navigating quickly between views.

aminomancer:
- Implement a basic cache for bug lists.
- When entering a view, first try to render from cached data, then start fetching from the server and update when data is received.
- Views show a subtle loading indicator when they have rendered from cached data but are currently fetching fresh data.
  - I figured this would be helpful since some views (like the triage view) take a very long time, and it would otherwise be hard to feel confident you're looking at the current state.
  - This doesn't interfere with displaying the normal loading indicator when no cached data was found for that view.
- Old caches can be invalidated by changing the version parameter in cache.js (there are ways to "expire" caches automatically - I will look into that and appreciate any advice)

EDIT: Btw, forgot to mention this also fixes the RS Messages view crash.

Fixes #188

### DIFF
--- a/src/bin/buildRange.ts
+++ b/src/bin/buildRange.ts
@@ -80,8 +80,8 @@ const REFERENCE_RELEASES: ReleaseDefinition[] = [
   },
   {
     startVersion: 98,
-    iterationsPattern: [2, 2]
-  }
+    iterationsPattern: [2, 2],
+  },
 ];
 
 export interface IterationLookup {
--- a/src/content/components/BugListView/BugListView.js
+++ b/src/content/components/BugListView/BugListView.js
@@ -2,14 +2,15 @@ import React from "react";
 import styles from "./BugListView.scss";
 import gStyles from "../../styles/gStyles.scss";
 import { BugList } from "../BugList/BugList";
-import { Loader } from "../Loader/Loader";
-import { runQuery } from "../../lib/utils";
+import { Loader, MiniLoader } from "../Loader/Loader";
+import { runQuery, matchQuery } from "../../lib/utils";
 
 export class BugListView extends React.PureComponent {
   constructor(props) {
     super(props);
     this.state = {
       loaded: false,
+      awaitingNetwork: false,
       bugs: [],
       query: {},
       uri: "",
@@ -19,6 +20,7 @@ export class BugListView extends React.PureComponent {
   }
 
   async componentWillMount() {
+    this._isMounted = true;
     const BASE_QUERY = {
       include_fields: this.props.columns.concat([
         "whiteboard",
@@ -29,15 +31,35 @@ export class BugListView extends React.PureComponent {
       ]),
       resolution: ["---", "FIXED"],
     };
-    const { bugs, query, uri } = await runQuery(
-      Object.assign({}, BASE_QUERY, this.props.query)
-    );
-    this.setState({
-      loaded: true,
-      bugs: this.props.sort ? bugs.sort(this.props.sort) : bugs,
-      query,
-      uri,
+    let query = Object.assign({}, BASE_QUERY, this.props.query);
+    await matchQuery(query)
+      .then(({ bugs, query, uri }) => {
+        if (this._isMounted) {
+          this.setState({
+            loaded: true,
+            awaitingNetwork: true,
+            bugs: this.props.sort ? bugs.sort(this.props.sort) : bugs,
+            query,
+            uri,
+          });
+        }
+      })
+      .catch(() => {});
+    await runQuery(query).then(({ bugs, query, uri }) => {
+      if (this._isMounted) {
+        this.setState({
+          loaded: true,
+          awaitingNetwork: false,
+          bugs: this.props.sort ? bugs.sort(this.props.sort) : bugs,
+          query,
+          uri,
+        });
+      }
     });
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
   }
 
   toggleDebug() {
@@ -74,6 +96,7 @@ export class BugListView extends React.PureComponent {
             {this.state.showDebug ? "Hide" : "Show"} Query
           </button>
         </p>
+        <MiniLoader hidden={!this.state.awaitingNetwork} />
         {this.state.showDebug ? this.renderDebug() : null}
       </React.Fragment>
     );
--- a/src/content/components/BugListView/BugListView.js
+++ b/src/content/components/BugListView/BugListView.js
@@ -3,7 +3,7 @@ import styles from "./BugListView.scss";
 import gStyles from "../../styles/gStyles.scss";
 import { BugList } from "../BugList/BugList";
 import { Loader, MiniLoader } from "../Loader/Loader";
-import { runQuery, matchQuery } from "../../lib/utils";
+import { runCachedQueries } from "../../lib/utils";
 
 export class BugListView extends React.PureComponent {
   constructor(props) {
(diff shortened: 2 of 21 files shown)


aminomancer:
This also fixes a bug where if you click one feature nav link, let it load for a while, then click another feature nav link before it's finished loading, it would render the first feature's bugs, at least until the second feature's bugs load. And it could actually permanently show the wrong bugs for a given feature if timed in such a way that the first feature takes longer to load than the second. It would basically go like this:
1. Click feature 1
2. Feature 1 starts loading (slow)
3. Click feature 2
4. Feature 2 starts loading (fast)
5. Feature 2 loads and renders
6. Feature 1 loads and renders
7. Idle (with feature 2's heading, but feature 1's bugs)

Here's an example (the location and heading are for "Messaging System" but the bugs are for "Feature Callout Tour for Firefox View")
render bug 1

The bug isn't really related to caching, but it's a lot easier to notice when caching is enabled, since some features can render way faster than others depending on whether they're already cached.

aminomancer:
@brennie for review

src/content/lib/utils.js
```diff
@@ -35,30 +36,70 @@ const FAKE_BUGS = [
...
+export async function runQueries(queries) {
+  return Promise.all(queries.map(runQuery));
+}
+
+export async function matchQuery(query) {
+  if (prefs.get("offline_debug")) {
+    return FAKE_BUGS;
+  }
+  if (prefs.get("disable_cache")) {
+    return Promise.reject();
```
brennie, line 79: Why not `throw new Error(...)` ?

src/content/lib/cache.js
```diff
@@ -0,0 +1,89 @@
...
+  constructor({ version }) {
+    this._version = version;
+  }
+
+  _open() {
+    if (prefs.get("disable_cache")) {
+      return null;
+    }
+    try {
+      const cache = caches.open(this._version);
```
brennie, line 13: maybe refer to `window.caches` or `globalThis.caches`?

src/content/lib/utils.js
```diff
@@ -35,30 +36,70 @@ const FAKE_BUGS = [
...
+    } catch (e) {
+      console.log("Error parsing cached response :>> ", response); // eslint-disable-line
+      console.log(query); // eslint-disable-line
+      console.error(e); // eslint-disable-line
+    }
+    if (data) {
+      return postProcess(data);
+    }
+  }
+  return data ? postProcess(data) : Promise.reject();
```
brennie, line 96: `data` can't be truthy here because we would have already returned.

`throw new Error(...)`

src/bin/buildRange.ts
```diff
@@ -80,8 +80,8 @@ const REFERENCE_RELEASES: ReleaseDefinition[] = [
   },
   {
     startVersion: 98,
-    iterationsPattern: [2, 2]
-  }
+    iterationsPattern: [2, 2],
+  },
```
brennie, line 84: Unrelated formatting changes -- put these in their own commit or drop them.

brennie (changes requested):
Some general notes:

- Can you clean up the commit history here? I'm not a fan of fixup commits (please ammend the commit instead of creating a new one) and some of your commit messages aren't clear or contain things like "WIP".
- I'm not a react expert by any means, but what I'm reading discourages `async componentWillMount` and suggests not to use `componentWillMount` at all.

src/content/components/BugListView/BugListView.js
```diff
@@ -29,17 +31,37 @@ export class BugListView extends React.PureComponent {
...
-    this.setState({
-      loaded: true,
-      bugs: this.props.sort ? bugs.sort(this.props.sort) : bugs,
-      query,
-      uri,
+    let query = Object.assign({}, BASE_QUERY, this.props.query);
+    await matchQuery(query)
+      .then(({ bugs, query, uri }) => {
+        if (this._isMounted) {
+          this.setState({
```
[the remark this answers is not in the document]
aminomancer: Yeah, that's the cause of the bug I mentioned in my 2nd comment in the PR

src/content/components/FeatureView/FeatureView.js
```diff
@@ -394,14 +400,18 @@ export class FeatureView extends React.PureComponent {
     return result;
   }
 
-  async getBugs(id) {
+  getBugs(id) {
```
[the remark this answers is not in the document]
aminomancer: I converted the awaits to callbacks since we don't want to wait for the cache get before starting the fetch (which is reliably much slower). So it doesn't need to be async anymore, and async functions are slightly slower due to the promise wrapping. But for the sake of consistency it might be worth keeping it async, since some of these bug view query methods do need to be async

src/content/lib/utils.js
aminomancer: Ah nice catch. In an earlier version I was using allSettled and basically expecting missing cache items so we could reuse old cache versions (e.g. from before a new meta was added or something). But the views aren't set up to be compatible with incomplete data yet. I might revisit that at some point but for now we'll just change the cache version any time we change a query set

src/content/components/BugListView/BugListView.js
```diff
@@ -19,6 +20,7 @@ export class BugListView extends React.PureComponent {
   }
 
   async componentWillMount() {
+    this._isMounted = true;
```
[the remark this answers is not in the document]
aminomancer: Will resolve with #191

src/content/components/Uplift/Uplift.js
```diff
@@ -36,45 +39,68 @@ export class Uplift extends React.PureComponent {
...
+                ["verified", "fixed"].includes(b.cf_status_beta)
+            ),
+          },
+        });
+      }
+    };
+    await matchQueries(queries)
+      .then(responses => updateState(responses, true))
+      .catch(() => {});
+    await runQueries(queries).then(responses => updateState(responses, false));
```
brennie, line 99: Same comment here re about making a `runCachedQueries()` method.

brennie (changes requested):

src/content/lib/utils.js
```diff
@@ -35,30 +36,70 @@ const FAKE_BUGS = [
...
       Accept: "application/json",
       "Content-Type": "application/json",
     },
-    body: JSON.stringify(query),
+    body,
     method: "POST",
   });
+  let fakeRequest = getFakeBugsRequest(body);
+  let resp = await fetch(request);
+  await cache.set(fakeRequest, resp);
```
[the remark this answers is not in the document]
aminomancer: I don't think we can abstract away the body string part, since the body can only be used once. So something like
```js
  let request = new Request("/api/bugs", {
    headers: {
      Accept: "application/json",
      "Content-Type": "application/json",
    },
    body,
    method: "POST",
  });
  let resp = await fetch(request);
  await cache.set(request, resp);
```
won't work. This would:
```js
  let request = new Request("/api/bugs", {
    headers: {
      Accept: "application/json",
      "Content-Type": "application/json",
    },
    body,
    method: "POST",
  });
  let resp = await fetch(request);
  await cache.set(body, resp);
```
But I guess it just feels a bit more opaque since it implies that we're keying the response to the body, when in reality we can only key responses to requests.

src/content/lib/cache.js
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      const cache = await this._open();
+      return cache && cache.match(request);
+    } catch (error) {
+      console.debug("Error accessing cache :>> ", error);
+      return null;
+    }
+  }
+
+  async set(request, response) {
```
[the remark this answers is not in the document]
aminomancer: `request.body` doesn't actually exist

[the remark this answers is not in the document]
brennie: We aren't pushing streaming bodies so we could do something like


```js
class CacheableRequest extends Request {
    constructor(url, options, body, method) {
        super(url, options, body, method);
        this.body = body;
    }

    asGetRequest() {
        return new Request(this.url, { headers: this.headers }, this.body, "GET");
    }
}
```

src/content/lib/cache.js
```diff
@@ -1,18 +1,14 @@
 import { prefs } from "./prefs";
 
-class Cache {
+class BugsCache {
```
aminomancer, line 3: seemed wise to rename this since `Cache` is the actual cache prototype lol

src/content/components/Loader/MiniLoader.scss
```diff
@@ -0,0 +1,100 @@
+/* Copyright (c) 2022 by Nobuaki Honma (https://codepen.io/nobuakihonma/pen/dYbqLQ)
```
brennie, line 1: nit: This should be a plain .css file.

brennie (approved):

AllegroFox (approved):
This looks great! Thank you so much for tackling it.

My only suggestion is more of an enhancement: it would be nice if the Miniloader scrolled with the page content, so that we could still see the status while scrolling a long list of bugs (particularly useful for CurrentIteration which is often really long.) 

We could do this by giving the Miniloader a height that's equal to the overflow content ([scrollHeight](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight) might be useful?) and then adding `position: sticky;` and a top margin (2em looked good to me) to the child <span/>. I tested it out by adding a giant height to the miniloader (2000px) and it looked nice, but it would be better if the height was dynamic since we don't know exactly how tall the list of loaded bugs will be and the effect won't work if the list is taller than the arbitrary height of the miniloader.

PR #206, opened 2022-11-18
Make the MiniLoader "sticky" so it maintains its position when user scrolls the main content

### LINKED ISSUE #204: Make the MiniLoader "sticky" so it maintains its position when user scrolls the main content
Per https://github.com/k88hudson/bugzy/pull/189#pullrequestreview-1179899937

> My only suggestion is more of an enhancement: it would be nice if the Miniloader scrolled with the page content, so that we could still see the status while scrolling a long list of bugs (particularly useful for CurrentIteration which is often really long.)
> 
> We could do this by giving the Miniloader a height that's equal to the overflow content ([scrollHeight](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight) might be useful?) and then adding `position: sticky;` and a top margin (2em looked good to me) to the child . I tested it out by adding a giant height to the miniloader (2000px) and it looked nice, but it would be better if the height was dynamic since we don't know exactly how tall the list of loaded bugs will be and the effect won't work if the list is taller than the arbitrary height of the miniloader.

I think we might be able to make a sticky container similar to `#toolbar` in Firefox's about:reader

aminomancer:
Fixes #204

### DIFF
--- a/src/content/components/Loader/MiniLoader.css
+++ b/src/content/components/Loader/MiniLoader.css
@@ -10,9 +10,9 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
   --loader-color: var(--ink-90);
   --load-ping-timing-function: cubic-bezier(0.075, 0.82, 0.165, 1);
   display: block;
-  position: absolute;
+  position: fixed;
   top: 0;
-  left: 0;
+  left: 220px;
   margin: 2em;
   height: 24px;
   width: 24px;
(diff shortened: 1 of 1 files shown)


aminomancer:
Lmk what you think of this - my thought process was basically scrollHeight can change between renders depending on some unpredictable factors, whereas the navbar is always 220px. But I can try it both ways

AllegroFox (approved):
Yep, this works beautifully! I somehow missed that the miniloader was a child the same as all the `<li>` when I played with it the first time, so allowing it to be relative to the parent is fine.

PR #208, opened 2022-11-18
Switch from node-sass to sass

### LINKED ISSUE #205: Investigate switching from node-sass to sass
node-sass requires node-gyp, which requires Python and also causes problems compiling on Windows. I think updating node and node-gyp might solve the Windows issue, and maybe will allow Python 3. But for what we're doing, it's probably not worth it, and node-sass is our only dependency that requires node-gyp. Removing it would greatly simplify this project and (we suspect) at very little cost. My understanding is [sass](https://www.npmjs.com/package/sass) is a bit slower, but we're compiling quite infrequently.

aminomancer:
Looks like this will be necessary for future deployments, since heroku's latest build stack [no longer supports python2](https://devcenter.heroku.com/articles/heroku-22-stack). For now I've postponed the upgrade so I could land the cache updates, but I'm not sure how long it'll let us do that. Switching looks safe though. It also massively increases the install speed and (at least for me on WSL) is actually faster on compile than node-sass was. But I'll look into whether it's possible to do some kind of dry run on heroku before trying to deploy these changes.

Also after merging this and before deploying, we must remember to go to the [heroku app settings](https://dashboard.heroku.com/apps/as-bugzy/settings) and click "Upgrade Stack" before pushing the master branch to heroku.

Fixes #205

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2061,12 +2061,6 @@
       "integrity": "sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=",
       "dev": true
     },
-    "amdefine": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz",
-      "integrity": "sha1-SlKCrBZHKek2Gbz9OtFR+BfOkfU=",
-      "dev": true
-    },
     "ansi-align": {
       "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


brennie (approved):
We may want to investigate moving to parcel v2 in the future as well.

aminomancer:
Successfully deployed 😄

PR #213, opened 2022-11-22
feat: Add patch and reviewer columns to bug lists

### LINKED ISSUE #187: Show status of attached patch(es) in current iteration view
As an individual contributor or engineering manager, it would helpful to see the state of work done on a given bugzilla ticket in the current iteration view. To keep things simple to start, this could simply be a column showing if there's at least one patch attached and whether it's a draft, in review, or approved to land.

aminomancer:
Fixes #187

Co-authored-by: yoenzhang <yozhang@mozilla.com>


src/server/queryUtils.ts
```diff
@@ -345,7 +346,8 @@ export async function fetchQuery(query: QueryConfig) {
 
   if (fetchAttachments) {
     let attachmentSets = bugs.map(bug => bug.attachments);
-    if (attachmentSets.filter(a => a).length > 0) {
+    attachmentSets = attachmentSets.filter(a => a.length);
```
aminomancer, line 349: I think `a => Array.isArray(a) && a.length > 0` is probably safer

aminomancer:
I see you fixed the "Edit all in Bugzilla" button showing up everywhere, so that's great. Here are the remaining issues I could think of, I might be missing something but we'll notice it again if so.
- 1200px max-width issue
- The colors need to be turned into CSS styles
- Investigate whether phabricator tells us which reviewers are blocking. We don't need to implement the logic or styles for this in this PR, but since you're now the most familiar with the Conduit API, you're best positioned to find the right method for getting this info.
- I mentioned before to investigate how it interacts with review groups. I just checked and it works fine. It turns them into tags. However, I would prefer if it added the hash mark `#` since that's the way you need to type it in commit messages. So if the reviewer object contains an indicator as to whether it's a group/project/tag, it'd be nice to mark that. Even if it's just a comment for now, that'll save us trouble trying to implement it in the future.

PR #233, opened 2023-01-21
feat: fetch iterations dynamically

### LINKED ISSUE #227: Instead of maintaining a hard-coded list of iterations, get the list from BMO
Using this query:
https://bugzilla.mozilla.org/rest/field/bug/cf_fx_iteration

aminomancer:
Fixes #227


aminomancer:
As of Jan 21, the dynamically retrieved iterations are nearly identical to the previous hard-coded iterations. The only differences seem to be errors in the old iterations. I set up a little tool to check all the iterations, you can uncomment it from [Preferences.js](https://github.com/mozilla/bugzy/blob/4890ca07e456a67b52ee045c7286507d9b2c5473/src/content/components/Preferences/Preferences.js) and then find it on the About Bugzy page. It just logs the iterations to the console

src/common/IterationLookup.ts
```diff
@@ -0,0 +1,280 @@
...
+ * 100.1) and a date range. Order is important so the date computations work
+ * correctly, and so we're able to parse duplicates as overrides.
+ * @example $ new IterationLookup(["101.1 - Jan 4 - 15", "101.2 - Jan 18 - 29"])
+ *          > {
+ *              byDate: { "2019-01-07": "101.1", "2019-01-14": "101.1", ... },
+ *              byVersionString: { "101.1": { startDate: "2019-01-04", ... } },
+ *              orderedVersionStrings: ["101.1", "101.2"], ...
+ *            }
+ */
+export class IterationLookup {
```
brennie, line 39: This is a class with no methods, only a ctor, which is a code smell.

This should be a function that returns a data structure.

Also having an interface `IterationLookup` and `class IterationLookup` is confusing from a typing perspective.

IMO rename this something like `buildIterationLookup` or `generateIterationLookup` or `lookupIterations.` and make it a function that returns an `IterationLookup` record.

src/common/IterationLookup.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  constructor(iterationLookup: IterationLookup) {
+    this.byVersionString = iterationLookup.byVersionString;
+    this.orderedVersionStrings = iterationLookup.orderedVersionStrings;
+    this.byDate = iterationLookup.byDate;
+    this.getWorkDays = Iterations.getWorkDays.bind(this);
+    this.getMondayBefore = Iterations.getMondayBefore.bind(this);
+    this.getIteration = this.getIteration.bind(this);
+    this.getAdjacentIteration = this.getAdjacentIteration.bind(this);
+  }
```
brennie, line 184: the IterationLookup class is only used by this, so I wonder if this ctor should be:

```
constructor(iterations: strings[]) {
  Object.assign(this, lookupIterations(iterations));

  // ...
}
```

src/common/IterationLookup.ts
```diff
@@ -0,0 +1,280 @@
...
+  getMondayBefore: (date: DateTime) => DateTime;
+
+  constructor(iterationLookup: IterationLookup) {
+    this.byVersionString = iterationLookup.byVersionString;
+    this.orderedVersionStrings = iterationLookup.orderedVersionStrings;
+    this.byDate = iterationLookup.byDate;
+    this.getWorkDays = Iterations.getWorkDays.bind(this);
+    this.getMondayBefore = Iterations.getMondayBefore.bind(this);
+    this.getIteration = this.getIteration.bind(this);
+    this.getAdjacentIteration = this.getAdjacentIteration.bind(this);
```
brennie, line 183: Do we need to explicitly bind these? We're not passing the methods as raw callbacks anywhere so this seems unnecessary.

src/common/IterationLookup.ts
```diff
@@ -0,0 +1,280 @@
...
+        .startOf("week")
+        .diff(start.startOf("week"), "weeks")
+        .toObject().weeks || 0;
+    let days = weeksBetween * DAYS_PER_WEEK;
+    const extraDays = Math.min(startWeekDay, 6) - 1;
+    const missingDays = Math.min(endWeekDay, 5);
+    // Remove/add extra days
+    days = days - extraDays;
+    days = days + missingDays;
+    return days;
```
brennie, line 255:
```
return days - extraDays + missingDays;
```

src/content/components/GlobalContext/GlobalContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  render() {
+    return (
+      <GlobalContext.Provider value={this.state}>
+        {this.props.children}
+      </GlobalContext.Provider>
+    );
+  }
+}
```
brennie, line 52: This can be a functional component.

```ts
export function GlobalContextProvider({metas, iterations, qm, children}: Readonly<GlobalContextProviderProps>) {
  const [metas, setMetas] = useState<MetaBug[]>(metas);
  const [iterations, setIterations] = useState<Iterations>(iterations);

  const value = useMemo(() => ({
    metas,
    iterations,
    setMetas,
    setIterations,
    qm,
  }, [metas, setMetas, iterations, setIteartions, qm]);

  return (<GlobalContext.Provider value={value}>{children}</GlobalContext.Provider>);
}
```

src/content/components/Preferences/Preferences.js
```diff
@@ -27,6 +27,12 @@ export class Preferences extends React.PureComponent {
     prefs.set(e.target.name, e.target.checked);
   }
 
+  // A button for testing iterations
+  async onIterationsCheck() {
+    const iterations = await (await fetch("/api/iterations")).json();
```
brennie, line 32: `await fetch().then(rsp => rsp.json())`

src/common/IterationLookup.ts
aminomancer: Thanks! Will fix these issues.

src/common/IterationLookup.ts
aminomancer: The server does the work for `lookupIterations` and caches it daily. Whereas this class takes an `IterationLookup` as input and outputs an `Iterations`. So consumers won't have the `iterations` array, just the `IterationLookup` object.

src/content/components/Preferences/Preferences.js
aminomancer: 😬 not sure where this pattern came from lol

src/content/components/GlobalContext/GlobalContext.tsx
aminomancer: Are you sure it doesn't cause everything to re-render every time the state changes? The reason I used a class component was because I read something about that. If you're not sure I can do some testing to find out.

brennie: State changes will cause re-renders in your class based component. If it didnt re-render, how would children get updated context?

aminomancer: Err not the context's own state, but the state of the context's parent. I found the article I was thinking of: https://reactjs.org/docs/context.html#caveats

src/content/components/GlobalContext/GlobalContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+  iterations: Iterations;
+  qm: QueryManager;
+}
+
+interface GlobalContextProviderProps extends GlobalContextProps {
+  children: React.ReactNode;
+}
+
+// Don't store any mutable state in here until React is updated to 18, consumers
+// will not reliably re-render on context changes.
```
aminomancer, line 25: I basically just put a warning here and punted off the solution to future me

aminomancer:
Should be up to date now. Complicated rebases

src/content/components/GlobalContext/GlobalContext.tsx
brennie: 👍

brennie (approved):
r+ with some minor nits.

src/common/IterationLookup.ts
```diff
@@ -0,0 +1,291 @@
...
+export class Iterations implements IterationLookup {
+  byDate: { [date: string]: string };
+  byVersionString: {
+    [versionString: string]: {
+      startDate: string;
+      endDate: string;
+      weeks: number;
+    };
+  };
+  orderedVersionStrings: string[];
```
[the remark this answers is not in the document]
aminomancer: wouldn't we need to change IterationLookup into a class? now it's just an interface returned by a regular factory function.

> 'IterationLookup' is not defined. eslint([no-undef](https://eslint.org/docs/rules/no-undef))
> Cannot extend an interface 'IterationLookup'. Did you mean 'implements'?ts(2689)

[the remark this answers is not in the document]
brennie: Ah my bad, I thought they were interchangeable. You can ignore that feedback then.

PR #235, opened 2023-01-22
feat: add an iteration picker

### LINKED ISSUE #232: Add an iteration picker
We support paths like `https://www.bugzy.org/iteration/:id`, but aside from the current and next iteration, the only way to access arbitrary iterations is by manually entering the path. We could add a <select> iteration picker in the navigation sidebar.

aminomancer:
Fixes #232

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -298,6 +298,18 @@
         "@babel/types": "^7.7.4"
       }
     },
+    "@babel/helper-string-parser": {
+      "version": "7.19.4",
+      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.19.4.tgz",
+      "integrity": "sha512-nHtDoQcuqFmwYNYPz3Rah5ph2p8PFeFCsZk9A/48dPc/rGocJ5J3hAAZ7pb76VWX3fZKu+uEr/FhH5jLx7umrw==",
+      "dev": true
+    },
+    "@babel/helper-validator-identifier": {
+      "version": "7.19.1",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.19.1.tgz",
+      "integrity": "sha512-awrNfaMtnHUr653GgGEs++LlAvW6w+DcPrOliSMXWCKo597CwL5Acf/wWdNkf/tfEQE3mjkeD1YOVZOUV/od1w==",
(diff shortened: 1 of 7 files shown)


aminomancer:
image

with dark reader:
image

aminomancer:
The searchbox and select dropdown menu

image

aminomancer:
I also tested it out with dates. Dates are definitely very useful, the only reason I'm uncertain about it is that they make the strings long enough to overflow. So in the <select>, the string gets truncated with an ellipsis, and in the dropdown, the <option>'s may wrap onto 2 lines.

image

aminomancer:
Could also switch to a numeric date format. It seems less legible though, and I have heard americans are more used to m/d/y, canadians more used to y/m/d, and everyone else uses d/m/y (or something like that).

nuggetsnegin (approved):
Tested and LGTM!

I don't find the truncation bad and agree having dates in that format as most legible. I think the only nitpick is the line break when the iteration string is too long when the menu is opened causing inconsistencies. But I don't think there's a good solution for this without hurting legibility (and it only seems to be a case with the EOY and current iteration).

image

aminomancer:
Yeahhh... it's a bit ugly having them wrap. Do you think showing the year is worthwhile? If it was the only thing wrapping I'd be inclined to get rid of the year display. But since the (current) wraps too, there wouldn't be much benefit. But maybe there's something else we can do instead of ` (current)`, like a star or something more concise. Any thoughts?

nuggetsnegin:
> Yeahhh... it's a bit ugly having them wrap. Do you think showing the year is worthwhile? If it was the only thing wrapping I'd be inclined to get rid of the year display. But since the (current) wraps too, there wouldn't be much benefit. But maybe there's something else we can do instead of ` (current)`, like a star or something more concise. Any thoughts?

Yeah I'm not sure if the year is worthwhile, I think it's obvious enough from the iteration # what year it should be. I like the idea of a star, just thinking out loud some other options could maybe be bolding the current iteration or an underline? I'm curious how the star would look tho, if it fits nicely that's probably most ideal.

aminomancer:
⁕
★
🟊
✦
✶
some unicode options

aminomancer:
Oh you know what... if we don't put (current) in the label, then you won't be able to search it up by typing `current`, which I think is a relatively useful feature. So I guess I'll just get rid of the years but not the (current) part.

PR #239, opened 2023-01-23
feat: add view for unassigned P1/P2 bugs

### LINKED ISSUE #219: Show prioritized bugs to take on
Maybe something like this query looking for unassigned P1/P2s:
https://bugzilla.mozilla.org/buglist.cgi?cmdtype=dorem&remaction=run&namedcmd=Messaging%20System%20Unassigned%20P1%2FP2&sharer_id=125983

It should probably be sorted by priority (P1 before P2) then iteration (current iteration then later then no iteration) then last updated. Unclear if we want to limit it to the first say 50 bugs vs all (currently 99).

This could probably be a separate view listed in the top left maybe just under Current Iteration?

cc @punamdahiya

rjfc:
Closes #219.

After some discussion with other members of the team, the view label has been set to "Unassigned P1/P2" although I am open to other ideas.

image


src/content/components/UnassignedView/UnassignedView.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+function sortBugs(bugs: any[]): any[] {
+  return bugs.sort((a, b) => {
+    let aIteration = cTrans.cf_fx_iteration(a.cf_fx_iteration);
+    let bIteration = cTrans.cf_fx_iteration(b.cf_fx_iteration);
+
+    let iterationComparison = 0;
+
+    if (aIteration != bIteration) {
+      if (aIteration === "--") {
```
aminomancer, line 89: I'm pretty sure the null value for iteration is `---` not `--`. I think two dashes is for priority. Can you confirm this?

aminomancer (changes requested):

src/content/components/UnassignedView/UnassignedView.tsx
rjfc: I am using the `cf_fx_iteration` function in `src/content/components/BugList/columnTransforms.js` to parse iteration values, which returns `--` as a null value:
```
cf_fx_iteration(value) {
  return parseIteration(value) || "--";
},
```

aminomancer: I see. That's just for UI. The parsing here is unnecessary, since this is only being done for sorting. Right now we're parsing every iteration string twice, once for sorting and again down the tree for rendering the text. But you can just compare the original iteration strings.
```js
parseFloat("110.1 - Jun 27 - 48") === 110.1;
parseFloat("---") === NaN;
parseFloat("") === NaN;
parseFloat() === NaN;
```

Also, bugs with no priority and no iteration are ending up at the top of the array. We have to check that they both actually have a priority before trying to compare them. This seems to work:
```js
[{cf_fx_iteration:NaN,priority:"P2",id:0},{cf_fx_iteration:"110.1",priority:"P2",id:1},{cf_fx_iteration:"110.1",priority:"P2",last_change_time:"2023-01-25T00:00:30.323Z",id:2},{cf_fx_iteration:"110.1",priority:"P2",last_change_time:"2023-01-25T00:00:31.423Z",id:3},{cf_fx_iteration:"110.1",priority:"P2",last_change_time:"2023-01-25T00:00:32.523Z",id:4},{cf_fx_iteration:"110.1",priority:"P2",last_change_time:"2023-01-25T00:00:33.931Z",id:5},{cf_fx_iteration:"110.2",priority:"P2",last_change_time:"2023-01-25T00:00:34.219Z",id:6},{cf_fx_iteration:NaN,priority:null,last_change_time:"2023-01-25T00:00:35.737Z",id:7},{cf_fx_iteration:"110.3",priority:"P1",last_change_time:"2023-01-25T00:00:36.152Z",id:8}]
.sort((a, b) => {
  let priorityComparison =
    a.priority && b.priority && a.priority.localeCompare(b.priority);
  if (priorityComparison) {
    return priorityComparison;
  }

  let iterationComparison = 0;
  if (a.cf_fx_iteration != b.cf_fx_iteration) {
    let aIteration = parseFloat(a.cf_fx_iteration);
    let bIteration = parseFloat(b.cf_fx_iteration);
    if (isNaN(aIteration)) {
      iterationComparison = Number(!isNaN(bIteration));
    } else if (isNaN(bIteration)) {
      iterationComparison = -1;
    } else {
      iterationComparison = aIteration - bIteration;
    }
  }

  return (
    iterationComparison ||
    Date.parse(b.last_change_time) - Date.parse(a.last_change_time)
  );
});
```

src/content/components/UnassignedView/UnassignedView.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { Container } from "../ui/Container/Container";
+import { Loader, MiniLoader } from "../Loader/Loader";
+import { columnTransforms as cTrans } from "../BugList/columnTransforms";
+
+interface GetQueryOptions {
+  iteration: string;
+  metas: Array<{
+    id: string;
+    component: string;
+    priority?: string;
```
nuggetsnegin, line 13: nitpick but wonder if this could be typed more tightly?

aminomancer: What do you have in mind? This is really a copy of MetaBug that appears everywhere, it's one of the types I had in mind for making a global declarations file. I exported it [here](https://github.com/mozilla/bugzy/pull/233/files#diff-e48222476c5f1fa0f5e7353b918bf31a422e62787ba7799d93d8739080124079) in my latest PR, though that hasn't landed yet

aminomancer: Btw this part is really just copied from IterationView.tsx I think. Most of this file is

rjfc: I'm realizing that I don't actually use this `metas` property in this view, so perhaps I can just remove it?

aminomancer: Yeah that's fine. Nice catch

aminomancer:
Hmm, you should roll back that commit. You have to rebase onto master, not add new commits. The former updates the parent so the diff contains only your changes. The latter just adds existing changes to yours, which creates conflicts rather than resolving them.

Just `git rebase master`

Or you can use GitLens, open the Branches subview in the Source Control view, make sure your PR branch is active, then right click the master branch > "Rebase current branch onto branch..." and when it asks if you want rebase or interactive rebase, you choose regular rebase.

aminomancer:
`git revert 8181968`

aminomancer:
Looks like you reverted, but you still need to rebase. I would switch to master, then pull, then switch back to your PR branch, then `git rebase master`

aminomancer (approved):
r+ thanks!

PR #240, opened 2023-01-26
Move types into declarations file

### LINKED ISSUE #234: Move types into declarations file(s)
The types are currently intermixed with implementation. I think it would be neater if we separated them, especially as we proceed with #126. Also I think in addition to declaration files for components, we should have one or two central declaration files, and consider moving some type declarations from narrow component files to this more central location. Because a lot of types are duplicated or will need to be imported across components. Basically put things like MetaBug or BugQuery in a central declarations file.

nuggetsnegin:
Fixes #234


aminomancer:
Did you mean to publish as draft?

src/types/global.types.ts
```diff
@@ -0,0 +1,89 @@
+interface Bug {
```
aminomancer, line 1: shouldn't the file be `global.d.ts`? I haven't seen `.types.ts` before

nuggetsnegin:
> Did you mean to publish as draft?

Ah yeah, forgot to publish as ready for review after the last commit. 😓

src/types/global.types.ts
nuggetsnegin: Oh you're right, I didn't realize that's the proper way to declare compile-time constants(?) will update!

nuggetsnegin: I wonder if I'm overpopulating the global types now that I think about it. 🤔 Based on [this](https://basarat.gitbook.io/typescript/project/modules/globals), wonder if it's best to use file based modules for most of these types? But I guess we don't have too many to work with right now.

aminomancer: Idk if there's really an issue with "polluting" the global namespace. And the goal for this issue is to share types across modules without needing imports. So if we didn't put them in a global declarations file, seems like we could just leave them where they are.

src/types/global.d.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+interface MetaBug {
+  id: string;
+  component?: string;
+  priority?: string;
+  displayName?: string;
+}
+
+// @: Is this type declaration necessary? Can we just use Bug?
```
aminomancer, line 12: Yup that would be fine

src/types/global.types.ts
aminomancer: let's ask someone with more TS experience @brennie

tsconfig.json
```diff
@@ -6,8 +6,7 @@
     "target": "ES2017",
     "outDir": "./dist",
     "allowJs": true,
-    "jsx": "react",
```
brennie, line 9: Why did you remove `"jsx": "react"` This is not mentioned in the PR.

src/types/global.types.ts
brennie: I find types being automatically imported from global.d.ts surprising tbh.

global.d.ts is, imo, for when you need to extend things already in the global namespace (HTML*Element, etc) to add custom properties, etc.

I would advise against putting all these types here.

brennie (commented):
You should also squash your fixup commits into your first commit to keep a clean history.

tsconfig.json
nuggetsnegin: Oh sorry, I should've left a comment or a separate commit but I noticed it was a duplicate import.

aminomancer:
Think we should just blow off adding a global declarations file? There is still some typing stuff that needs adjustment, though it's not gonna be fully functional till we update some node modules. You can test how it actually works by running `node_modules/.bin/tsc -p tsconfig.json --noEmit` which you can see we need to update typescript and react. But you could still clean up the types a bit I imagine

nuggetsnegin:
> Think we should just blow off adding a global declarations file? There is still some typing stuff that needs adjustment, though it's not gonna be fully functional till we update some node modules. You can test how it actually works by running `node_modules/.bin/tsc -p tsconfig.json --noEmit` which you can see we need to update typescript and react. But you could still clean up the types a bit I imagine

Gotcha, sorry for delay I de-prioritized this for feature work. Will fix this PR up next week! 👍🏻

PR #243, opened 2023-02-07
Export API endpoint in project_settings.js and fix Prefs and Store system 

### LINKED ISSUE #242: Add a setting for testing on bugzilla staging instance
> I just noticed the test bugs in Bugzilla. Just in case you didn't know, there's a staging instance of Bugzilla at https://bugzilla-dev.allizom.org/ - If possible, it would be better to test on that rather than on the production instance.

_Originally posted by @Standard8 in https://github.com/mozilla/bugzy/issues/238#issuecomment-1413310305_
            

We have a file `prefs.js` with some debugging/development settings. You can add a new pref there that flips the origin from `https://bugzilla.mozilla.org` to `https://bugzilla-dev.allizom.org`. I think there are a couple files with API URL related definitions, so you may want to consolidate anything that references `https://bugzilla.mozilla.org` into a single exported symbol. And then just have those various files import `ROOT_URL` or whatever, where `ROOT_URL` is defined in relation to the new setting.

To avoid rate limits you may need to create a new account and API key for this instance (the API key is specified as an environment variable, so for testing on the BMO instance you can just use your own personal key from the BMO settings page, same with testing on the staging instance). So if you do run into rate limits, you can ask for someone to help set an account up in the [#bmo](https://mozilla.slack.com/archives/C5F7KFJ4R) slack channel. Any other questions just ask me on slack.

rjfc:
- Replace all instances of Mozilla API endpoint with exported symbol in `project_settings.js`
- Fix Store and Prefs system to set default values

Closes #242.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2206,6 +2206,22 @@
                 "readable-stream": "^2.0.6"
               }
             },
+            "balanced-match": {
+              "version": "1.0.0",
+              "bundled": true,
+              "dev": true,
+              "optional": true
+            },
+            "brace-expansion": {
+              "version": "1.1.11",
+              "bundled": true,
+              "dev": true,
+              "optional": true,
+              "requires": {
+                "balanced-match": "^1.0.0",
+                "concat-map": "0.0.1"
+              }
+            },
             "chownr": {
               "version": "1.1.3",
               "bundled": true,
@@ -2214,6 +2230,12 @@
             },
             "code-point-at": {
               "version": "1.1.0",
+              "bundled": true,
+              "dev": true,
+              "optional": true
+            },
+            "concat-map": {
+              "version": "0.0.1",
               "bundled": true,
               "dev": true,
               "optional": true
@@ -2342,6 +2364,12 @@
               "dev": true,
               "optional": true
             },
+            "ini": {
+              "version": "1.3.5",
+              "bundled": true,
+              "dev": true,
+              "optional": true
+            },
             "is-fullwidth-code-point": {
               "version": "1.0.0",
               "bundled": true,
@@ -2356,6 +2384,15 @@
               "bundled": true,
               "dev": true,
               "optional": true
+            },
+            "minimatch": {
+              "version": "3.0.4",
+              "bundled": true,
+              "dev": true,
+              "optional": true,
+              "requires": {
+                "brace-expansion": "^1.1.7"
+              }
             },
             "minimist": {
               "version": "0.0.8",
@@ -5682,14 +5719,6 @@
             "function-bind": "^1.1.1"
           }
         },
-        "loose-envify": {
-          "version": "1.4.0",
-          "resolved": "https://registry.npmjs.org/loose-envify/-/loose-envify-1.4.0.tgz",
-          "integrity": "sha512-lyuxPGr/Wfhrlem2CL/UcnUc1zcqKAImBDzukY7Y5F/yQiNdko6+fRLevlw1HgMySw7f611UIY408EtxRSoK3Q==",
-          "requires": {
-            "js-tokens": "^3.0.0 || ^4.0.0"
-          }
-        },
         "resolve": {
           "version": "1.17.0",
           "resolved": "https://registry.npmjs.org/resolve/-/resolve-1.17.0.tgz",
@@ -7395,7 +7424,8 @@
     "js-tokens": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/js-tokens/-/js-tokens-3.0.2.tgz",
-      "integrity": "sha1-mGbfOVECEw449/mWvOtlRDIJwls="
+      "integrity": "sha1-mGbfOVECEw449/mWvOtlRDIJwls=",
+      "dev": true
     },
     "js-yaml": {
       "version": "3.13.1",
--- a/src/common/IterationLookup.ts
+++ b/src/common/IterationLookup.ts
@@ -283,7 +283,7 @@ export class Iterations implements IterationLookup {
    * (or today if no date is given).
    * @param {number} diff n for upcoming iterations, -n for previous iterations
    * @param {string|DateTime} [dateString] defaults to today
-   * @returns {LegacyIteration}
+   * @returns {LegacyIteration|null}
    */
   getAdjacentIteration(
     diff: number,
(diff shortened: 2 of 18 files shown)


src/content/lib/prefs.js
```diff
@@ -1,13 +1,23 @@
 // const Store = window.require('electron-store');
-import prefDefaults from "../../config/pref_defaults";
+const prefDefaults = {
+  // The user's bugzilla email
```
rjfc, line 3: `prefDefaults` was moved from the `pref_defaults.js` file (which was then deleted) and into `prefs.js` since that is the only place where `prefDefaults` is referenced.

src/common/IterationLookup.ts
```diff
@@ -251,7 +251,9 @@ export class Iterations implements IterationLookup {
         ? DateTime.fromISO(dateString)
         : dateString;
     const monday = getMondayBefore(date);
-    const iterationString = this.byDate[monday.toISODate()];
+    const iterationString =
+      this.byDate[monday.toISODate()] ||
+      Object.values(this.byDate)[Object.values(this.byDate).length - 1];
```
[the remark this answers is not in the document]
aminomancer: This is a generic method so it should return an undefined iteration if one doesn't exist. For example, this confounds the behavior of getAdjacentIteration. It's too cryptic for this to return valid iterations for invalid inputs. Handling this case should be done by the callers. Instead, you should just add a safety check.

```js
const iterationString = this.byDate[monday.toISODate()];
const iteration = this.byVersionString[iterationString];
return iteration ? {
  number: iterationString,
  start: iteration.startDate,
  due: iteration.endDate,
} : null;
```

and then change the jsdoc comment at the top so it says
```js
 * @returns {LegacyIteration|null}
```

and then add another method `getLatestIteration`

```js
/**
 * Get the latest iteration.
 * @returns {LegacyIteration}
 */
getLatestIteration(): LegacyIteration {
  const iterationString = this.orderedVersionStrings[
    this.orderedVersionStrings.length - 1
  ];
  const iteration = this.byVersionString[iterationString];
  return {
    number: iterationString,
    start: iteration && iteration.startDate,
    due: iteration && iteration.endDate,
  };
}
```

so the consumers can use `iterations.getIteration() ?? iterations.getLatestIteration()`

Only the UI consumers, though. Like, `getAdjacentIteration` should not be changed.

aminomancer (changes requested):
- fix the iteration methods

src/common/IterationLookup.ts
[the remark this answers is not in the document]
rjfc: I updated `getAdjacentIteration` to return `null` if `this.getIteration(dateString)` is null. Is that fine?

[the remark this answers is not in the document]
rjfc: Also, I use `getAdjacentIteration(-1, this.context.iterations.getLatestIteration().start)` for the current iteration rather than `getLatestIteration()` so I don't invalidate parts of Bugzy that refer to the "next iteration" in which case I use `getLatestIteration()`

aminomancer:
In the future, you should probably amend previous commits rather than adding new ones. It gets a bit unwieldy with so many tiny commits, especially if all you're doing is refactoring. In this case we'll have to squash them when we merge

package-lock.json
```diff
@@ -2206,6 +2206,22 @@
                 "readable-stream": "^2.0.6"
```
aminomancer, line 2206: You haven't updated any node modules, so package-lock.json should not be modified. These are probably changes to your local environment that you made in previous revisions and didn't revert

aminomancer: In the future, at least on this repo, if you're not modifying package.json, you should always discard changes to package-lock.json before committing. Now it's a bit more annoying to restore after you've already committed the changes.

src/common/IterationLookup.ts
```diff
@@ -265,21 +283,23 @@ export class Iterations implements IterationLookup {
    * (or today if no date is given).
    * @param {number} diff n for upcoming iterations, -n for previous iterations
    * @param {string|DateTime} [dateString] defaults to today
-   * @returns {LegacyIteration}
+   * @returns {LegacyIteration|null}
    */
   getAdjacentIteration(
     diff: number,
     dateString?: string | DateTime
   ): LegacyIteration {
```
aminomancer, line 291: actually, you need to change the return types in the signature here too. it should say
```suggestion
  ): LegacyIteration | null {
```

src/common/IterationLookup.ts
```diff
@@ -242,7 +242,7 @@ export class Iterations implements IterationLookup {
    * For a given date (or no date for today), return the iteration number, the
    * start date, and the due date.
    * @param {string|DateTime} [dateString] defaults to today
-   * @returns {LegacyIteration}
+   * @returns {LegacyIteration|null}
    */
   getIteration(dateString?: string | DateTime): LegacyIteration {
```
aminomancer, line 247: same here
```suggestion
  getIteration(dateString?: string | DateTime): LegacyIteration | null {
```

src/common/IterationLookup.ts
```diff
@@ -265,21 +283,23 @@ export class Iterations implements IterationLookup {
...
    * @param {string|DateTime} [dateString] defaults to today
-   * @returns {LegacyIteration}
+   * @returns {LegacyIteration|null}
    */
   getAdjacentIteration(
     diff: number,
     dateString?: string | DateTime
   ): LegacyIteration {
-    const baseIteration: string = this.getIteration(dateString).number;
+    const baseIteration: string = this.getIteration(dateString)?.number;
```
aminomancer, line 292: and this should be
```suggestion
    const baseIteration: string | void = this.getIteration(dateString)?.number;
```

src/content/components/Exports/Exports.js
```diff
@@ -5,6 +5,8 @@ import { GlobalContext } from "../GlobalContext/GlobalContext";
 import { BugList } from "../BugList/BugList";
 import { Loader, MiniLoader } from "../Loader/Loader";
 import { DateTime } from "luxon";
+import { prefs } from "../../lib/prefs";
```
aminomancer, line 8: why did you add this import?

src/content/components/FeatureList/FeatureList.js
```diff
@@ -97,7 +98,13 @@ export class FeatureList extends React.PureComponent {
 
   render() {
     const release = Number(
-      this.context.iterations.getIteration().number.split(".")[0]
+      this.context.iterations.getIteration()?.number.split(".")[0] ??
+        this.context.iterations
+          .getAdjacentIteration(
```
aminomancer, line 103: not sure what's going on here, this seems to be getting the second-latest iteration, not the latest iteration

PR #245, opened 2023-02-07
fix: use default values in prefs/store system

### LINKED ISSUE #244: fix: the prefs/store system
The prefs/store system does not currently make use of the default values. We need to fix it so that it sets prefs to their default values in localStorage if they haven't been set yet.

rjfc:
Closes #244.

### DIFF
--- a/src/content/lib/prefs.js
+++ b/src/content/lib/prefs.js
@@ -1,21 +1,32 @@
 // const Store = window.require('electron-store');
-import prefDefaults from "../../config/pref_defaults";
+const prefDefaults = {
+  // The user's bugzilla email
+  bugzilla_email: "",
+  // Disable all client-side network requests
+  offline_debug: false,
+  // Disable the user cache for bug views
+  disable_cache: false,
+};
 
 class Store {
-  get(key) {
-    return localStorage.getItem(key);
+  get(pref, def) {
+    return (
+      JSON.parse(globalThis.localStorage?.getItem(pref)) ??
+      def ??
+      prefDefaults[pref]
+    );
   }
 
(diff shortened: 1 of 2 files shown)


src/content/lib/prefs.js
```diff
@@ -1,13 +1,23 @@
...
+  // Disable the user cache for bug views
+  disable_cache: false,
+};
 
 class Store {
-  get(key) {
-    return localStorage.getItem(key);
+  get(key, defaultValue) {
+    if (localStorage.getItem(key) === null && key in prefDefaults) {
+      this.set(key, defaultValue);
```
aminomancer, line 14: see my comment in the issue, the get method should not have any side effects

src/content/lib/prefs.js
```diff
@@ -1,13 +1,23 @@
...
+    }
+    return JSON.parse(localStorage.getItem(key));
```
aminomancer, line 16: btw not sure why we need JSON here. all the prefs are already valid JSON types

rjfc: localStorage stores values as strings. For `offline_debug` and `disable_cache`, we want a boolean `false`, not a string `"false"` (which happens without this JSON.parse)

aminomancer: Makes sense, we just don't need JSON.stringify

aminomancer: well I guess if user plugs in `"true"` to bugzilla_email it could bork something, so go ahead and keep it

aminomancer (approved):
Lgtm thanks!

PR #246, opened 2023-02-14
feat: add options for setting bug priority and iteration

rjfc:


src/content/components/BugList/BugList.js
```diff
@@ -93,32 +162,223 @@ export class BugList extends React.PureComponent {
...
+      );
+
+      if (selection.value == 0) {
+        const newIteration =
+          selectedBug.cf_fx_iteration == "---"
+            ? this.context.iterations.getIteration().number
+            : this.context.iterations.getAdjacentIteration(
+                1,
+                null,
+                lookupIterations([selectedBug.cf_fx_iteration])
```
aminomancer, line 200: This method `lookupIterations` is for the server, not content. It is generated once per day and requested from the server by `/api/iterations` (see server.ts). The content then generates an IterationLookup from it, which is then passed down to all the components via React context. See other examples that use `iterations` from GlobalContext.

src/common/IterationLookup.ts
```diff
@@ -265,13 +270,18 @@ export class Iterations implements IterationLookup {
...
    */
   getAdjacentIteration(
     diff: number,
-    dateString?: string | DateTime
+    dateString?: string | DateTime,
+    baseIterationString?: string
   ): LegacyIteration {
-    const baseIteration: string = this.getIteration(dateString).number;
+    if (!dateString) dateString = DateTime.local();
+    const baseIteration: string = !baseIterationString
```
aminomancer, line 282: logic is clunky here, should be inverted

src/content/components/BugList/BugList.js
```diff
@@ -14,18 +44,52 @@ const EditorGroup = props => (
...
+      editSelection: -1,
+      newPriority: "---",
+      bugUpdating: false,
     };
     this.onCheck = this.onCheck.bind(this);
     this.onAllSelectedCheck = this.onAllSelectedCheck.bind(this);
     this.onCheckShowResolved = this.onCheckShowResolved.bind(this);
   }
 
+  componentWillMount() {
```
aminomancer, line 87: Isn't this method called `UNSAFE_componentWillMount` in React 18?

aminomancer (changes requested):

src/content/components/BugList/BugList.js
```diff
@@ -261,4 +519,5 @@ BugList.defaultProps = {
   crossOutResolved: true,
   showResolved: true,
   visibleIfEmpty: true,
+  fetchBugs: () => {},
```
rjfc, line 522: I added a prop `fetchBugs` to `BugList` which is a function that is called when a bug's priority or iteration is updated which should fetch the bugs again to receive the new iteration or priority of the updated bug. 

In `OtherView.tsx`, `IterationView.tsx`, and `UnassignedView.tsx`, the `useBugFetcher` hook is used to retrieve bugs. [`useBugFetcher`](https://github.com/mozilla/bugzy/blob/master/src/content/hooks/useBugFetcher.ts#L86) takes a parameter for items to add to its `useEffect` dependency array (`updateOn`), which I could make use of. But I'm not too sure what the best is to translate a button click event to a variable to be placed in a `useEffect` dependency array.

What are your thoughts on the best way to re-trigger these `useBugFetcher`s from within `BugList`? The other files don't use `useBugFetcher` so I can just pass [a function that fetches bugs](https://github.com/mozilla/bugzy/blob/set-bug-iteration-and-priority/src/content/components/FeatureView/FeatureView.js#L595-L597) into `BugList`, but I can't do that here.

src/content/components/BugList/BugList.js
```diff
@@ -93,32 +162,223 @@ export class BugList extends React.PureComponent {
...
+  onUpdateIterationSelect(selection) {
+    this.setState({ newIteration: selection.value });
+  }
+
+  onUpdatePrioritySelect(selection) {
+    this.setState({ newPriority: selection.value });
+  }
+
+  updateBug(id, fields) {
+    this.setState({ bugUpdating: true });
```
rjfc, line 221: `bugUpdating: true` renders a text in place of the edit dropdown that says "Updating bug...". Right now, after updating the iteration or priority, it displays the "Updating bug..." text, the list refreshes, then after about 1-2 seconds the bug's iteration or priority is updated. I believe that this 1-2 second delay is because we first return cached bug data before returning bug data that is fetched from Bugzilla. Is this delay okay, or should I find an alternative to accessing cached bug data first?

src/content/components/BugListView/BugListView.js
```diff
@@ -79,6 +101,9 @@ export class BugListView extends React.PureComponent {
           tags={true}
           bugs={this.state.bugs}
           columns={this.props.columns}
+          fetchBugs={() => {
+            this.fetchBugs();
+          }}
```
aminomancer, line 106: this is going to recreate the function on every render. just use `fetchBugs={this.fetchBugs}`, and then in the constructor add `this.fetchBugs = this.fetchBugs.bind(this)`

in functional components, you can do `const fetchBugs = React.useCallback(() => {})` and pass that instead

src/content/components/BugList/BugList.js
```diff
@@ -38,25 +37,20 @@ const selectStyle = {
...
+    ...provided,
+    color: "hsl(0, 0%, 20%)",
+  }),
 };
 
-function getDisplayName(id) {
+const getDisplayName = id => {
   return definitions[id] ? definitions[id].displayName : id;
-}
+};
```
aminomancer, line 48: there's no reason to change this, it's not a component

src/content/components/BugList/BugList.js
```diff
@@ -80,41 +74,64 @@ const priorityOptions = [
...
-export class BugList extends React.PureComponent {
-  constructor(props) {
-    super(props);
-    this.state = {
-      selectedBugs: {},
-      showResolved: props.showResolved,
-      editSelection: -1,
-      newIteration: currentIteration,
-      newPriority: "---",
+export const BugList = ({ ...props }) => {
```
aminomancer, line 77: `{ ...props }` doesn't make sense. just use `props` or else spread out all the main props like `{ showResolved, bugs, ...props }`

src/content/components/BugList/BugList.js
```diff
@@ -80,41 +74,64 @@ const priorityOptions = [
...
+  const handleMenuOpen = useCallback(() => {
+    let { current } = iterationSelectRef;
+
+    requestAnimationFrame(() => {
+      current.focusedOptionRef?.scrollIntoView({ block: "nearest" });
+    });
+  }, []);
 
-  getRowClassName(bug) {
+  const getRowClassName = bug => {
```
aminomancer, line 115: same here, keep the traditional `function () {}` syntax unless it's a function component (i.e. it returns JSX)

aminomancer: also, this should be moved out of BugList. the way you've changed this, you're redefining all of these functions one very render. the content of a functional component is executed on every render. that's why you'd use `useCallback`. but for these kinds of things, it's better to define it outside of the component so React doesn't need to read over it every time. just change the parameters or something.

really, changing this to a functional component is not such a great idea to do in this PR. it's already becoming quite unwieldy to review. you should try to keep changes as minimal as possible, just as a general rule. I think we'd be better off making these general refactorings in a separate PR.

src/content/components/BugList/BugList.js
```diff
@@ -80,41 +74,64 @@ const priorityOptions = [
...
+      bug.assigned_to === props.bugzilla_email
     ) {
       classNames.push(styles.mine);
     }
     return classNames.join(" ");
-  }
+  };
 
-  renderColumn(columnId, bug) {
+  const renderColumn = (columnId, bug) => {
```
aminomancer, line 130: these functions that return JSX should be their own component, if they're not part of a class component. the reason to use a class component would be that you don't want to waste resources re-defining this on every render. you could avoid that by making it a new component, but this will be a lot easier in general if you just leave it as a class component

aminomancer:
I'm guessing the reason you changed BugList to a functional component is so you could use `React.useContext`? But you updated React to 18.2.0... which means the bug with class component context is fixed. So, you can just add a static property to class components `static contextType = GlobalContext;` and then reference the context as `this.context`. See BugListView for example

src/content/components/Triage/Triage.js
```diff
@@ -116,6 +116,65 @@ export class Triage extends React.PureComponent {
     this._isMounted = false;
   }
 
+  async fetchBugs() {
+    const prevIteration = this.context.iterations.getAdjacentIteration(-1)
+      .number;
+    const result = await this.context.qm.runQueries([
```
aminomancer, line 122: If all these queries are the same as they are in `componentWillMount`, then you should try to deduplicate them. Move the queries into a class method called `getQueries` that just returns the array that is passed to runQueries here. Then all the methods that use the queries (fetchBugs and componentWillMount) can just do `this.context.qm.runQueries(this.getQueries())`

src/content/components/Uplift/Uplift.js
```diff
@@ -20,6 +20,69 @@ export class Uplift extends React.PureComponent {
...
+        ]),
+        component: BUGZILLA_TRIAGE_COMPONENTS,
+        order: "changeddate DESC",
+        custom: {
+          [betakey]: { anyexact: ["?", "+", "blocking"] },
+          "flagtypes.name": { notsubstring: `approval-mozilla-beta` },
+        },
+      },
+    ]);
+    this.setState({
```
aminomancer, line 62: same kind of situation here, this state transformation could be moved into a method. then both fetchBugs and componentWillMount can share it, so we won't need to duplicate all this text, and I'll be able to read the whole PR without having a brain aneurysm 😂

src/common/IterationLookup.ts
```diff
@@ -265,13 +270,18 @@ export class Iterations implements IterationLookup {
    * (or today if no date is given).
    * @param {number} diff n for upcoming iterations, -n for previous iterations
    * @param {string|DateTime} [dateString] defaults to today
+   * @param {string} [baseIterationString] optionally provided iteration rather than date
```
aminomancer, line 273: this line is too long. max line for comments is 80. download [this VS code extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) and bind its command to a keyboard shortcut like Cmd+Shift+R. then in your VS code settings JSON, add these:
```json
  "rewrap.wrappingColumn": 80,
  "editor.rulers": [80],
```
that will add a vertical ruler on the 80th column so you can clearly see when a comment is overflowing the limit, and it will set the limit to 80 for the rewrap extension.

you might also find this useful, if you use [an extension](https://marketplace.visualstudio.com/items?itemName=phoihos.git-commit-message-editor) to write commit messages:
```json
  "[git-commit]": {
    "rewrap.wrappingColumn": 72,
    "rewrap.autoWrap.enabled": true,
    "editor.rulers": [50, 72]
  }
```
that will give you rulers at 50 and 72 for commit messages specifically. which is useful since those are the limits for git commit messages - 50 for the first line (the short message) and 72 for subsequent lines (the extended description)

rjfc: Thanks, that's super useful!

PR #251, opened 2023-03-09
feat: add invalid routes and error page

### LINKED ISSUE #218: Add something like a 404 not found page for invalid routes
In triage today we weren't able to change iterations since the current iteration doesn't exist on BMO yet, so we marked bugs by the previous iteration. But there's no nav link for the previous iteration, so in order to reach these bugs we need to enter /iteration/109.2 and navigate manually. I guess there might be other cases where we have occasion to do that. I accidentally just typed in /109.2 and realized the page is simply blank with no error logged or anything. So, I think we should have some kind of no-route handling

rjfc:
To test the invalid routes page, navigate to any nonexistent route:
image

To test the error page, turn off your internet and load Bugzy, or replace `fetch("/api/metas")` or `fetch("/api/metas")` on lines 11 and 16 in `src/content/index.tsx` with an invalid URL:
image

Closes #218.

### DIFF
--- a/src/content/components/ActiveRSMessages/ActiveRSMessages.tsx
+++ b/src/content/components/ActiveRSMessages/ActiveRSMessages.tsx
@@ -3,6 +3,8 @@ import { getTargetingAttributes } from "./TargetingParser";
 import styles from "./ActiveRSMessages.scss";
 import { columnTransforms } from "../BugList/columnTransforms";
 import { fetchBugById, RSMessage } from "../../../server/queryUtils";
+import ReactDOM from "react-dom";
+import { ErrorView } from "../ErrorView/ErrorView";
 
 interface CFRMessage extends RSMessage {
   content: {
@@ -149,7 +151,17 @@ export class ActiveRSMessages extends React.PureComponent {
       })
       .then(messages =>
         this.setState({ messages, selectedBucket: BUCKETS[bkey] })
-      );
+      )
+      .catch(() => {
+        ReactDOM.render(
+          <ErrorView
+            header={"Error"}
+            subheader={"There was an error fetching data."}
+            buttonText={"Try again"}
+          />,
+          document.getElementById("root")
+        );
+      });
   }
 
   bucketUpdate(el: React.ChangeEvent<HTMLSelectElement>) {
--- a/src/content/components/ErrorView/ErrorView.tsx
+++ b/src/content/components/ErrorView/ErrorView.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import gStyles from "../../styles/gStyles.scss";
+
+type ErrorViewProps = {
+  header: string;
+  subheader: string;
+  buttonText: string;
+  buttonHref?: string;
+};
+
+const containerStyle = {
+  maxWidth: 900,
+  margin: "0 auto",
+  padding: "20px 40px",
+  textAlign: "center",
+};
+
+const refreshPage = () => {
+  window.location.reload();
+};
+
+export class ErrorView extends React.PureComponent<ErrorViewProps> {
+  render() {
+    return (
+      <div style={containerStyle}>
+        <h1>{this.props.header}</h1>
+        <p>{this.props.subheader}</p>
+        <a
+          className={gStyles.primaryButton}
+          onClick={!this.props.buttonHref ? refreshPage : null}
+          href={this.props.buttonHref || ""}>
+          {this.props.buttonText}
+        </a>
+      </div>
+    );
+  }
+}
--- a/src/content/components/ErrorView/ErrorView.tsx
+++ b/src/content/components/ErrorView/ErrorView.tsx
@@ -12,7 +12,7 @@ const containerStyle = {
   maxWidth: 900,
   margin: "0 auto",
   padding: "20px 40px",
-  textAlign: "center",
+  textAlign: "center" as const,
 };
 
 const refreshPage = () => {
--- a/src/content/components/Preferences/Preferences.js
+++ b/src/content/components/Preferences/Preferences.js
@@ -1,0 +1,110 @@
+import React from "react";
+import styles from "./Preferences.scss";
+import { prefs } from "../../lib/prefs";
+import ReactDOM from "react-dom";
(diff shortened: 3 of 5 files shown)


rjfc:
How important is it for the error page to have the navbar on the left and priority guide on the right? Is it worth modifying the `Router` component to take some props so that I can just do `ReactDOM.render(<Router props={..} />, document.getElementById("root"));` and force `Router` to render the `ErrorView` page on each `fetch`'s `catch`? Currently, I'm just doing `ReactDOM.render(<ErrorView />, document.getElementById("root"));`.

aminomancer:
For the invalid route case, can you add a "Go Back" button that calls history.back()?
And then center the paragraph that says "The page you were looking for was not found." since it looks a little weird in left alignment.

As for navbar on the error page, if the internet is not working, you're not gonna be able to download any of this stuff. If you had the bugzy files in http cache, then theoretically the navbar could load, but the metas will not work since the metas cache is on the server, not the client. So, I don't really see any point in showing any content on the error page aside from a standard 404. This looks perfect to me

aminomancer (changes requested):
-

aminomancer:
Actually maybe not "Go Back" since this may be the first bugzy page in history. So just do "Current Iteration" and have it navigate to current_iteration

rjfc:
> Actually maybe not "Go Back" since this may be the first bugzy page in history. So just do "Current Iteration" and have it navigate to current_iteration

Done

src/content/index.tsx
```diff
@@ -4,20 +4,32 @@ import { GlobalContextProvider } from "./components/GlobalContext/GlobalContext"
...
+      .then(res => res.json())
+      .catch(() => {
+        ReactDOM.render(<ErrorView />, document.getElementById("root"));
+      }),
+    fetch("/api/iterations")
+      .then(res => res.json())
+      .catch(() => {
+        ReactDOM.render(<ErrorView />, document.getElementById("root"));
+      }),
   ]);
```
aminomancer, line 21: Can't you `catch` on the `Promise.all` call?

aminomancer (changes requested):

src/content/index.tsx
rjfc: Good idea!

aminomancer:
Cool, can you remind me to review these PRs next week?

aminomancer:
suggested changes:
```diff
diff --git a/src/content/components/Router/Router.js b/src/content/components/Router/Router.js
index 066583e..730dfc0 100644
--- a/src/content/components/Router/Router.js
+++ b/src/content/components/Router/Router.js
@@ -114,6 +114,9 @@ const RouterNav = withRouter(
             {routes
               .filter(route => !route.hidden)
               .map((route, i) => {
+                if (route.hide) {
+                  return null;
+                }
                 if (route.spacer) {
                   return <li key={i} className={styles.spacer} />;
                 } else if (route.header) {
@@ -490,8 +493,7 @@ export class Router extends React.PureComponent {
         },
       },
       {
-        label: "404",
-        icon: "info",
+        hide: true,
         routeProps: {
           path: "*",
           render: () => (
```

src/content/components/ErrorView/ErrorView.tsx
```diff
@@ -0,0 +1,37 @@
...
+  subheader: string;
+  buttonText: string;
+  buttonHref?: string;
+};
+
+const containerStyle = {
+  maxWidth: 900,
+  margin: "0 auto",
+  padding: "20px 40px",
+  textAlign: "center",
```
aminomancer, line 15:
```ts
  textAlign: "center" as const,
```

src/content/components/Preferences/Preferences.js
```diff
@@ -1,6 +1,8 @@
 import React from "react";
```
aminomancer, line 1: Conflict with #263 - needs to be rebased and revised to work w/ your new settings page

src/content/index.tsx
```diff
@@ -4,20 +4,35 @@ import { GlobalContextProvider } from "./components/GlobalContext/GlobalContext"
...
-    </GlobalContextProvider>,
-    document.getElementById("root")
-  );
+  try {
+    const [metas, iterationsLookup] = await Promise.all([
+      fetch("/api/metas").then(res => res.json()),
+      fetch("/api/iterations").then(res => res.json()),
+    ]);
+
+    if (metas && iterationsLookup) {
```
aminomancer, line 16: Why is there a condition here? If the condition is not met, this function will do nothing, without throwing, and nothing will be rendered. I think we want this to throw so the error page would be shown?

src/content/components/ActiveRSMessages/ActiveRSMessages.tsx
```diff
@@ -149,7 +151,17 @@ export class ActiveRSMessages extends React.PureComponent {
...
+      )
+      .catch(() => {
+        ReactDOM.render(
+          <ErrorView
+            header={"Error"}
+            subheader={"There was an error fetching data."}
+            buttonText={"Try again"}
+          />,
+          document.getElementById("root")
+        );
```
aminomancer, line 163: Hmmm... subcomponents shouldn't be taking over rendering. Really, `ReactDOM.render` should only ever be called once. Ideally, if there's a catastrophic error, the subcomponent should navigate via the router. One way to do that would be setting an `errored` state, and if true, the render method should render a [Redirect](https://v5.reactrouter.com/web/api/Redirect) to the error page.

Though it will take some effort and research on the react router docs to determine the best way to pass the necessary parameters to the error route (like what location the try again button should navigate to). I'm sure there are lots of blog posts and discussions on how to do an error page with react router, so we could figure it out.

This also requires testing to make sure we don't break the site, so I would be fine with just blowing off this issue for now. These bugs (and the string bugs they each depend on) are much higher priority:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1828170
- https://bugzilla.mozilla.org/show_bug.cgi?id=1828172

aminomancer:
fixed by 5502226cc35f9121ba3f0b2c5f7980739757408b

PR #252, opened 2023-03-10
feat: add year to iteration picker

### LINKED ISSUE #237: Compute iteration dates and expose them to UI
For the dropdown in #235, and in other iteration-related views, we might want to show years in iteration date strings. Between 67 and 120, we cover several years.

rjfc:
Closes #237.

### DIFF
--- a/src/content/components/IterationPicker/IterationPicker.tsx
+++ b/src/content/components/IterationPicker/IterationPicker.tsx
@@ -29,8 +29,12 @@ export const IterationPicker: React.FunctionComponent<IterationPickerProps> = ({
     (iteration: string) => {
       let string = iteration;
       const { start, due } = context.iterations.getDatesForIteration(iteration);
-      let startString = start.toFormat("LLL d");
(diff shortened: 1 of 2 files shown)


rjfc:
Does this format look odd?
image

aminomancer:
hmm. what percent of items do you think wrap onto 2 lines?

aminomancer:
and no actually it looks pretty good IMO.

maybe we should make the sidebar wider. I've noticed many meta bug titles get cut off. and it's not like we need the space in the main body. can you just change that in this bug? making the sidebar wider should automatically make the iteration picker wider.

aminomancer:
see what's the smallest width you can make the sidebar so that none of the route labels get cut off

rjfc:
> hmm. what percent of items do you think wrap onto 2 lines?

Before increasing the sidebar width, about 50%.

> see what's the smallest width you can make the sidebar so that none of the NavLinks get cut off

Done:
image

aminomancer (approved):

PR #254, opened 2023-03-27
chore: use images instead of text for Material Icons

### LINKED ISSUE #253: Use images instead of text for Material Icons
We use Material Icons for the bug type icons in our bug lists (specifically, the defect/enhancement/task icons). But we use the implementation where the element's text content is just rendered as an icon using the Material icons typeface.

Which is nice in a lot of ways, but the main thing I dislike about it is it's still text, it's selectable, hovering it shows the text cursor, etc. There are ways to avoid that, like `.material-icons{user-select:none}`, but that applies even if the font failed to load.

And that's another problem with the implementation. It relies on the user not disabling web fonts and Google's servers working, or on the user having Material Icons installed for some reason, and it also relies on the user not overriding fonts at the browser or system levels.

And in the end, I think we only need 3 icons anyway. So just serving our own SVG or PNG files would also be more lightweight. It'd be good to set up anyway in case we add more icons in the future, or want to mix and match icons from other sources with material icons.

So, this would involve changing the components that render icons via Material Icons, adding icon files, and adding CSS rules to display the icon files.

How to display the icons? The advantage of using `` tags is you get alt text and better a11y semantics. That seems best to me, but a blank element with `background-image` works too. I think the main reason I'd use CSS for this would be to avoid the ugly "broken" placeholder icon when the icons fail to load. Seems unlikely the icons would fail to load, but if they did, `background-image` would make it effectively blank. It wouldn't be able to convey any information since there would be no alt text. So I think img tags are probably best, but if you can think of a good reason to do something else, that's fine.

rjfc:
Closes #253.


src/content/components/BugList/BugList.scss
```diff
@@ -210,19 +210,7 @@ $monospace: SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New"
   }
 }
 
-.task, .defect, .enhancement {
+.bugIcon {
   vertical-align: top;
   margin-top: -2px;
```
aminomancer, line 215: I wonder if we still need these rules now. Never looked too closely at this, but at a glance it seems like these might have been intended to compensate for the typesetting weirdness. Can you look into that?

src/content/img/icons/defect.svg
aminomancer, line 1: Pretty sure these invisible paths can be improved
```svg
<path d="M0 0h24v24H0z" fill="none"/>
```
And there are redundant fill attributes on both the svg and path elements. You only need it on one. As these are all monotone icons, I would advise putting the fill attribute on the svg element, and removing it from the path element. So, the path only needs one attribute, `d`

aminomancer (changes requested):

src/content/img/icons/defect.svg
aminomancer: Ah I remember. These are little invisible squares with the same dimensions as the viewbox. Not sure why Google includes them. I think it is probably for some illustration applications that don't understand viewbox/artbox so when they import it, they end up clipping the viewbox to the dimensions of the paths. So if the path's bounding box is smaller than the viewbox, it would get clipped if there isn't a hidden path somewhere in the file with bounding box equal to the viewbox.

But that's not necessary for our purposes, so it can just be removed.

src/content/components/BugList/BugList.scss
aminomancer: i.e. does removing these affect the vertical centering of the icons relative to the text? we might still want to keep them but worth checking

src/content/components/BugList/columnTransforms.js
aminomancer, line 1: Do this instead:

```diff
diff --git a/src/content/components/BugList/columnTransforms.js b/src/content/components/BugList/columnTransforms.js
index 58f0091..9e708fd 100644
--- a/src/content/components/BugList/columnTransforms.js
+++ b/src/content/components/BugList/columnTransforms.js
@@ -4,9 +4,7 @@ import tagConfig from "../../../config/tags";
 import styles from "./BugList.scss";
 import priorityStyles from "../PriorityGuide/PriorityGuide.scss";
 import { DateTime } from "luxon";
-import taskIcon from "../../img/icons/task.svg";
-import enhancementIcon from "../../img/icons/enhancement.svg";
-import defectIcon from "../../img/icons/defect.svg";
+import icons from "../../img/icons/*.svg";

 const OPEN_BUG_URL = "https://bugzilla.mozilla.org/show_bug.cgi?id=";

@@ -127,19 +125,7 @@ export const columnTransforms = {
     return getShortName(value);
   },
   type(value) {
-    let icon = "";
-    switch (value) {
-      case "task":
-        icon = <img src={taskIcon} />;
-        break;
-      case "defect":
-        icon = <img src={defectIcon} />;
-        break;
-      case "enhancement":
-        icon = <img src={enhancementIcon} />;
-        break;
-    }
-    return icon;
+    return <img src={icons[value]} alt={value} title={value} />;
   },
   cf_fx_points(value) {
     return value === "---" ? "" : value;
```

Adds alt text and tooltip. And more concise

aminomancer (changes requested):

PR #256, opened 2023-03-28
feat: turn icons in priority column into buttons that open the priority guide

### LINKED ISSUE #255: Turn icons in priority column into buttons that open the priority guide
The icons in the priority column of the bug list are currently plain text. They should be turned into buttons that open the Priority Guide on the right. The arrow button in the priority guide should also have a `cursor: pointer` and a hover state.

The `--` should also be horizontally aligned with the rest of the priority icons in the column:
image

rjfc:
This PR allows the user to click on the icons in the priority column to open the priority guide on the right. It also adds a `cursor: pointer` when the priority guide button is hovered.

https://user-images.githubusercontent.com/17807246/228357286-39bfd1a5-cba5-4f29-9374-53de7d9d3183.mov

This PR also aligns the no priority text ("--") with the other priority icons.
image

Closes #255.


src/content/components/BugList/columnTransforms.js
```diff
@@ -145,9 +145,17 @@ export const columnTransforms = {
...
+  priority(priority, _, props) {
+    function openPriorityGuide() {
+      props.priorityGuide.current.openPriorityGuide();
+    }
+
     return (
-      <span className={priorityStyles[priority.toLowerCase()]}>{priority}</span>
+      <button
+        className={priorityStyles[priority.toLowerCase()]}
+        onClick={priority == "--" ? null : openPriorityGuide.bind(this)}>
```
aminomancer, line 156: Are you sure we need to pass so many props around? This is the kind of thing I added the GlobalContext for. It might make sense to make a component called PrefObserver that stores pref values, gets them from localStorage at startup, mirrors them in localStorage, and allows stuff to subscribe to changes. This is kinda hard to explain so let me try to draft something like that and then you can see if it works for this

aminomancer: See my PR #259

rjfc: I don't think this will work since localStorage stores data in strings. I need to store a React Ref of PriorityGuide so that I can call one of its functions from another component.

aminomancer: It works fine. I already tested it last week. The prefs storage stringifies and parses values automatically. Wasn't it you who added that feature?

You just need to add a subscription callback in the PriorityGuide constructor or on mounted. Just like I did for the other components in #259.

```js
prefs.on("priorityGuideCollapsed", this.onPrefChange);
...
  onPrefChange({ name, newValue } = {}) {
    if (name === "priorityGuideCollapsed") {
      this.setState({ priorityGuideCollapsed: newValue });
    }
  }
```

Then your buttons don't need to do anything except
```js
prefs.set("priorityGuideCollapsed", false);
```

aminomancer: Btw, the pref name should be `priority_guide_collapsed` or something. Since all the other pref names are already snake cased, not camel cased

aminomancer:
Did you rebase onto #259 already? That has to be merged into master first, or the diff will include it in the changes instead of in the parent revision.

src/content/components/BugList/columnTransforms.js
rjfc: Thanks! I was misunderstanding something about your suggestion previously

aminomancer:
I need you to review it first. Then after it merges into master, you rebase this PR onto master. You should also squash all the commits down after that, since it seems like some of my commits got in the history here somehow

src/content/components/BugList/columnTransforms.js
```diff
@@ -146,8 +147,16 @@ export const columnTransforms = {
...
   priority(priority) {
+    function openPriorityGuide() {
+      prefs.set("priority_guide_collapsed", false);
+    }
+
     return (
-      <span className={priorityStyles[priority.toLowerCase()]}>{priority}</span>
+      <button
+        className={priorityStyles[priority.toLowerCase()]}
+        onClick={priority == "--" ? null : openPriorityGuide.bind(this)}>
```
aminomancer, line 157: No need for `bind` since it doesn't use `this`

The reason you would not use an arrow function in the prop value is to avoid re-creating the function on every render. But this doesn't avoid that, since the function statement is inside the render method, and isn't protected behind a `useCallback` hook or anything. This isn't a functional React component, so the hook isn't an option...

So basically, `function openPriorityGuide` belongs outside of columnTransforms. It should just be defined up, outside of this block.

Btw, I actually don't like `priority_guide_collapsed` - it seems inverted. Can you change it to `priority_guide_open` and just flip its default value to true?

src/content/components/PriorityGuide/PriorityGuide.js
```diff
@@ -1,17 +1,29 @@
 import React from "react";
 import styles from "./PriorityGuide.scss";
+import { prefs } from "../../lib/prefs";
 
 export class PriorityGuide extends React.PureComponent {
   constructor(props) {
     super(props);
-    this.state = { priorityGuideCollapsed: false };
+    this.onPrefChange = this.onPrefChange.bind(this);
     this.onToggleCollapse = this.onToggleCollapse.bind(this);
```
aminomancer, line 9: This method no longer references `this` so no need to bind it

src/content/components/PriorityGuide/PriorityGuide.scss
```diff
@@ -56,6 +64,10 @@
   @media(max-width: 1210px) {
     display: none;
   }
+
+  &:hover {
+    cursor: pointer;
+  }
```
aminomancer, line 70: Doesn't need to go in `:hover` - cursor will only apply when it's hovered anyway. So this actually just makes it a tiny bit slower, needlessly. I'd just put it in the rules in the block above

src/content/components/PriorityGuide/PriorityGuide.js
rjfc: It references `this` for `this.setState({ priorityGuideCollapsed: newValue });`

aminomancer: That's onPrefChange, not onToggleCollapse. But now that you've changed onToggleCollapse to use `this.state` it does need to be bound, so my comment isn't applicable anymore

aminomancer:
I think you're missing a default value for `priority_guide_open`. It should be `true` so the guide starts out open. I think the way it's set up now, since there's no default value, the pref value is effectively null, so the guide starts out closed.

src/content/components/PriorityGuide/PriorityGuide.scss
```diff
@@ -23,6 +23,10 @@
   padding: 0 3px;
   border-radius: 3px;
   font-weight: bold;
+
+  &:hover {
+    cursor: pointer;
+  }
```
aminomancer, line 29: You don't need to put this under `:hover`. It can go with the other properties above. The `cursor` property can only ever do anything if the cursor is over the element (in other words, when the element is hovered)

aminomancer (changes requested):
Mostly looks good, just a couple issues

src/content/lib/prefs.js
```diff
@@ -6,25 +6,32 @@ const prefDefaults = {
...
 class Store {
   get(pref, def) {
     let storedValue = globalThis.localStorage?.getItem(pref);
     let value;
     try {
-      value = JSON.parse(storedValue ?? "null");
+      value = JSON.parse(storedValue);
+      if (value == null) {
+        throw null;
+      }
```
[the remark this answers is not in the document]
rjfc:
```
value = value ?? def ?? prefDefaults[pref];
if (storedValue) {
  this.set(pref, value);
}
```
was inside the `catch` block, so I added the `if (value == null) { ... }` to trigger it even if `value` is null. But I guess that might have been an accident and it makes more sense to just move that out of the `catch` block, thanks!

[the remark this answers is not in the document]
aminomancer: That's not quite what I meant. The purpose of putting `set` in the `get` method is only for migrating invalid prefs. After this last change, we're rewriting to storage every time we get a non-nullish value from storage. We only want to write to storage in order to correct values that are invalid JSON.

So I think this is the correct form:

```js
  get(pref, def) {
    let storedValue = globalThis.localStorage?.getItem(pref);
    let value;
    try {
      value = JSON.parse(storedValue ?? "null");
    } catch (error) {
      // If the stored value is not valid JSON, try to fix it.
      switch (typeof (def ?? prefDefaults[pref])) {
        case "boolean":
          value = storedValue === "true";
          break;
        case "number":
          value = Number(storedValue);
          break;
        default:
          value = storedValue;
      }
      if (storedValue) {
        this.set(pref, value ?? def ?? prefDefaults[pref]);
      }
    }
    return value ?? def ?? prefDefaults[pref];
  }
```

aminomancer (approved):
Nice, thanks! Pleasure working w/ you 🥂

PR #263, opened 2023-04-17
feat: add settings page

### LINKED ISSUE #262: Add a proper Settings page
Existing settings are currently exposed only in the About page, and disabled in production. It'll be easier to add future settings if we have a regular Settings page for them. In the navigation sidebar it would probably go just above the About Bugzy nav link. Could probably just copy the settings table JSX out of the About component and put it in the new SettingsView component.

rjfc:
Closes #262.


src/content/components/SettingsView/SettingsView.js
```diff
@@ -0,0 +1,93 @@
...
+
+  onPrefChange({ name, newValue } = {}) {
+    const newState = {};
+    newState[name] = newValue;
+    this.setState(newState);
+  }
+
+  // A button for testing iterations
+  onIterationsCheck() {
+    fetch()
```
aminomancer, line 37: While we're here, can you fix this?
```js
fetch("/api/iterations")
...
```
Not sure how I screwed this up lol

src/content/components/SettingsView/SettingsView.js
```diff
@@ -0,0 +1,93 @@
...
+                value={this.state.bugzilla_email}
+              />
+            </div>
+          </div>
+          <div className={styles.row}>
+            <div className={styles.col}>
+              <label htmlFor="offline_debug">
+                Debug in offline mode (fake data)
+              </label>
+            </div>
```
aminomancer, line 72: Btw, would be good to add a checkbox for `disable_cache` if you have time.

aminomancer (approved):
Looks good, thanks!

aminomancer:
Oh shit, hang on

src/content/components/SettingsView/SettingsView.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // A button for testing iterations
+  onIterationsCheck() {
+    fetch()
+      .then(rsp => rsp.json())
+      .then(data => console.log(data));
+  }
+
+  componentWillMount() {
+    this.state = {
```
aminomancer, line 43: Should be `this.setState({...})`

Not your fault, this was already here, but good opportunity to fix it

aminomancer: Actually I'm probably the one who copied this lol

src/content/components/Router/Router.js
```diff
@@ -480,15 +483,26 @@ export class Router extends React.PureComponent {
...
         icon: "info",
         routeProps: {
           path: "/about",
-          component: Preferences,
+          component: AboutView,
         },
       },
-    ];
+    ].filter(
+      router => !router.devOnly || process.env.NODE_ENV !== "production"
```
aminomancer, line 504: Filtering is done around line 115. You can just add a new condition in the filter function up there for this, to return null. No reason to actually exclude this route for production, just exclude the list item. That way we can still use it by entering the URL directly

aminomancer: And better to avoid looping over all of these twice

src/content/components/Router/Router.scss
```diff
@@ -104,3 +104,6 @@
 .icon-hourglass {
   background-image: url("../../img/icons/hourglass.svg");
 }
+.icon-settings {
+  background-image: url("../../img/icons/settings.svg");
+}
```
aminomancer, line 109: Should be a newline here

Issue #271, opened 2023-08-03
Add an OMC Tickets view


aminomancer:
Related to #270. This view should basically start by querying all the unresolved bugs in our components that have a "See Also" field whose value contains the substring `mozilla-hub.atlassian.net/browse/OMC`. Then it should loop over each of the returned bugs to organize them by ticket, then make a BugList for each ticket.

```js
let tickets = {};
for (let bug of bugs) {
  let ticket = bug.seeAlso.value.match(
    /mozilla-hub.atlassian.net\/browse\/(OMC-\d+)/
  );
  if (ticket) {
    tickets[ticket[1]] = tickets[ticket[1]] || [];
    tickets[ticket[1]].push(bug);
  }
}

let lists = [];
for (let ticket in tickets) {
  lists.push(
    <BugList
      bugs={tickets[ticket]}
      title={
        <a href={`https://mozilla-hub.atlassian.net/browse/${ticket}`}>
          {ticket}
        </a>
      }
    />
  );
}
```

And each BugList should look basically like this, sorted by points, then by priority.

### [OMC-123]()

| \[check all\] | 3 bugs \| 13 points |     |                |          |           |          |        |
| ------------- | ------------------- | :-: | -------------- | -------- | --------- | -------- | ------ |
|               | Bug                 |     | Title          | Assignee | Iteration | Priority | Points |
| [ ]           | [1234 567]()        | ➕  | Lorem ipsum    | yozhang  | 118.1     | P2       | 7      |
| [ ]           | [2345 678]()        | 📋  | dolor sit amet | shughes  | 118.1     | P1       | 3      |
| [ ]           | [3456 789]()        |  ☀  | consectetur    | pdahiya  | 118.1     | P1       | 3      |

The BugLists themselves should be sorted by total points. So I think you could basically do

```js
lists.sort((a, b) => {
  return (
    b.props.bugs.reduce((acc, bug) => acc + bug.points.value, 0) -
    a.props.bugs.reduce((acc, bug) => acc + bug.points.value, 0)
  );
});
```

PR #272, opened 2023-08-03
Add an Engineering Allocation view

### LINKED ISSUE #270: Add an Engineering Allocation view
Add a view that includes an individual BugList for each OMC engineer, tracking all unresolved bugs in our components that are assigned to that engineer and have an iteration whose first part (e.g. the 117 in 117.1) matches the current iteration. It should look kind of like the table below. Results should be sorted by points first, then by priority. In other words, 7-pointers go before 3-pointers, but if there are two 7-pointers, then the P1 should go before the P2.

### shughes
| \[check all\] | 3 bugs \| 13 points |              |     |                |           |          |        |
| ------------- | ------------------------------ | ------------ | :-: | -------------- | --------- | -------- | ------ |
|               | Ticket                         | Bug          |     | Title          | Iteration | Priority | Points |
| [ ]           | [OMC-123]()                    | [1234 567]() | ➕  | Lorem ipsum    | 118.1     | P2       | 7      |
| [ ]           | [OMC-456]()                    | [2345 678]() | 📋  | dolor sit amet | 118.1     | P1       | 3      |
| [ ]           | [OMC-789]()                    | [3456 789]() |  ☀  | consectetur    | 118.1     | P1       | 3      |

aminomancer:
Fixes #270


aminomancer:
Could add more bells and whistles, but this should work for #270 for now. You can also use this for reference for #271, I think. Similar queries, but just processing the bugs differently.

src/content/components/AllocationView/AllocationView.js
```diff
@@ -0,0 +1,277 @@
...
+import { columnTransforms } from "../BugList/columnTransforms";
+
+// Add a view that includes an individual BugList for each OMC engineer,
+// tracking all unresolved bugs in our components that are assigned to that
+// engineer, have an iteration matching the current release, and have the points
+// field set. Display columns should include ticket, bug id, type, title,
+// iteration, priority, and points, in that order. Bugs should be sorted by
+// points first, then by priority. In other words, 7-pointers go before
+// 3-pointers, but if there are two 7-pointers, then the P1 should go before the
+// P2.
```
aminomancer, line 18: These longer notes are just for development to make the spec easier to find. I'll remove em when we're done

src/content/img/icons/capacity.svg
aminomancer, line 1: I should really reverse the colors here, bigger is badder so the big square should be red and the small square green. I can do that before we land it

yoenzhang (approved):
Looks good! Tested on local, everything seems to work as intended

aminomancer:
Awesome, thanks for checking. I'll fix up my comments and merge

PR #275, opened 2023-08-09
Jira View

yoenzhang:


src/content/components/JiraView/JiraView.js
```diff
@@ -0,0 +1,205 @@
...
+        {
+          include_fields: includeFields,
+          resolution: "---",
+          order: "cf_fx_points DESC, priority, changeddate DESC",
+          rules: [
+            {
+              key: "assigned_to",
+              operator: "anyexact",
+              value: teams.omc.join(","),
+            },
```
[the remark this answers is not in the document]
aminomancer: But keep in mind an `OR` rule can't be the first in the array so you need to change the order to put a normal rule first.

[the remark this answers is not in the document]
aminomancer: I would put this first in place of `assigned_to`, to exclude meta bugs
```js
            { key: "keywords", operator: "notsubstring", value: "meta" },
```

src/content/components/JiraView/JiraView.js
```diff
@@ -0,0 +1,205 @@
...
+                  operator: "anyexact",
+                  value: BUGZILLA_TRIAGE_COMPONENTS.join(","),
+                },
+              ],
+            },
+            {
+              key: "cf_fx_iteration",
+              operator: "substring",
+              value: this.props.release,
+            },
```
aminomancer, line 115: You're missing the requirement:
> that have a "See Also" field whose value contains the substring mozilla-hub.atlassian.net/browse/OMC

```js
            {
              key: "see_also",
              operator: "substring",
              value: "mozilla-hub.atlassian.net/browse/OMC-",
            },
```

This seems to make the query faster.

src/content/components/JiraView/JiraView.js
```diff
@@ -0,0 +1,205 @@
...
+              if (ticket) {
+                tickets[ticket[1]] = tickets[ticket[1]] || [];
+                tickets[ticket[1]].push(bug);
+              }
+            }
+          }
+        }
+
+        for (const ticket in tickets) {
+          // if there are no bugs for this engineer, remove them from the list
```
aminomancer, line 137:
```suggestion
        for (const ticket of Object.keys(tickets)) {
          // if there are no bugs for this ticket, remove it
```

src/content/components/JiraView/JiraView.js
```diff
@@ -0,0 +1,205 @@
...
+    return (
+      <Container
+        loaded={true}
+        heading={"Jira Tickets"}
+        subHeading={
+          <React.Fragment>
+            This list includes unresolved, sized bugs that are assigned to a
+            Jira ticket.
+          </React.Fragment>
+        }>
```
aminomancer, line 172:
```suggestion
        subHeading="This list includes unresolved OMC bugs that are assigned to a Jira ticket.">
```

aminomancer (changes requested):

aminomancer:
Can you add "fixes #271" to the extended commit message?

package-lock.json
aminomancer, line 1: There's still a lockfile change. Are you not using lts/fermium? Should be easy to switch back and forth with nvm. There's an nvmrc file in the repo so it should be automatic.

src/content/components/JiraView/JiraView.js
```diff
@@ -0,0 +1,201 @@
...
+              }
+            }
+          }
+        }
+
+        for (const ticket of Object.keys(allJiraTickets)) {
+          if (!allJiraTickets[ticket].length) {
+            delete allJiraTickets[ticket];
+            continue;
+          }
```
aminomancer, line 140: I think this isn't necessary for JiraView, since you only add a ticket to allJiraTickets `if (ticket)`. It was only necessary for AllocationView because we added engineers to the list by copying from `teams.omc` rather than by looping over bugs.

src/content/components/Router/Router.js
```diff
@@ -322,6 +323,15 @@ export class Router extends React.PureComponent {
           render: props => <AllocationView {...props} release={release} />,
         },
       },
+      {
```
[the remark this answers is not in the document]
aminomancer: This is still below AllocationView, I would prefer it be above AllocationView

aminomancer (changes requested):
Looks great, just some final cleanup.

When you merge, make sure to choose "squash & merge," since you have multiple commits in this PR.

yoenzhang:
Removed the package-lock.json files, it should align with those on master, completed the other requested changes, should be good to squash and merge

aminomancer (changes requested):
still below allocation view

aminomancer (approved):
Thanks! Make sure to choose "squash & merge"

PR #278, opened 2023-08-20
Upgrade to parcel v2 and react 19, v2

### LINKED ISSUE #211: Update to parcel v2?

aminomancer:
Fixes #211


aminomancer:
@yoenzhang Let me know if you can build this successfully. The build process is very simple and much faster with this PR.

```
git fetch origin

git checkout -b aminomancer/issue211

npm i

npm start
```

Should take just about 1 second now.

src/content/components/ActiveRSMessages/ActiveRSMessages.tsx
```diff
@@ -1,8 +1,18 @@
 import * as React from "react";
 import { getTargetingAttributes } from "./TargetingParser";
-import styles from "./ActiveRSMessages.scss";
+import * as styles from "./ActiveRSMessages.module.scss";
 import { columnTransforms } from "../BugList/columnTransforms";
-import { fetchBugById, RSMessage } from "../../../server/queryUtils";
```
aminomancer, line 5: This frontend code was previously importing server code! 🤯

aminomancer:
I set engines.node to 19, as it seems a bit faster on 19, but it actually works with 18.

aminomancer:
If anyone has 5 minutes to approve this, I'd greatly appreciate it. I know it's too large to read through all of the bulk changes. What would help the most would be simply testing it on macOS or Linux, whatever you mainly use. Test steps are really fast:

```
git fetch origin

git checkout -b aminomancer/issue211

npm i

npm start
```

Then open `localhost:1989` (if 1989 is taken, set the `PORT` env var to something free and it'll use that port instead) in your browser, and click a few pages. If something isn't working you should see a crash as soon as one of the views loads, or you'd probably even get an exception before you can open the page.

I'll just bypass the branch protection otherwise, as I need to merge this PR to continue Bugzy work on my main device. `npm install` no longer works for me, maybe because I upgraded to mozillabuild 4.1 and I have to put its python version on my path for other things to work correctly.

Parcel v1 requires deasync, but that fails on windows if you're using python 3.11, and I really don't want to micromanage my global python version in addition to my node version. If we have to put effort into this, that effort is better spent on future-proofing this tool, rather than pouring more time and effort into ancient software. So simply updating to the non-broken parcel v2 is the ideal solution.

nuggetsnegin:
Hm, I'm getting:
```
Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
    main index.tsx:28
    AsyncFunctionThrow self-hosted:856
    (Async: async)
    parcelRequire<["index.tsx"]< index.tsx:53
    newRequire content.f69400ca.js:47
    parcelRequire content.f69400ca.js:81
    <anonymous> content.f69400ca.js:120
```

after running npm i, npm start

aminomancer:
> Hm, I'm getting:
> 
> ```
> Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
>     main index.tsx:28
>     AsyncFunctionThrow self-hosted:856
>     (Async: async)
>     parcelRequire<["index.tsx"]< index.tsx:53
>     newRequire content.f69400ca.js:47
>     parcelRequire content.f69400ca.js:81
>     <anonymous> content.f69400ca.js:120
> ```
> 
> after running npm i, npm start

Do you have any existing .cache or .parcel-cache files?

nuggetsnegin (approved):
We synced up over this on slack, it's working for me locally now! R+

PR #281, opened 2023-10-17
Add the useless but charming "cosmic rays" tag

AllegroFox:
Dan's mention of the possible effects of cosmic rays during our retro was so delightful that I felt the need to add a tag we could use, on any weird telemetry bugs or intermittents we can’t explain. Because joy, and no other reason.

Appears when the [cosmic rays] whiteboard tag is used on Bugzilla.


aminomancer:
Lmao that's great. Nice one. Thanks.

aminomancer (approved):

PR #285, opened 2023-12-07
fix: update sizing to use fibonacci sequence.

### LINKED ISSUE #284: Update sizing to use Fibonacci sequence

aminomancer:
Fixes #284

### DIFF
--- a/src/content/components/BugList/columnTransforms.js
+++ b/src/content/components/BugList/columnTransforms.js
@@ -214,11 +214,11 @@ export const columnTransforms = {
 
     let tShirtSize;
     let points = parseInt(value, 10);
-    if (points <= 3) {
+    if (points < 3) {
       tShirtSize = "SM";
-    } else if (points <= 7) {
+    } else if (points < 8) {
       tShirtSize = "MD";
-    } else if (points <= 12) {
+    } else if (points < 15) {
       tShirtSize = "LG";
     } else {
       tShirtSize = "XL";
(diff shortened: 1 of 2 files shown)


aminomancer:
- Ron said we could add back XL but give it 20+ for potential future use. Didn't check if Bugzilla has points that go that high, I'll probably just give it the max number Bugzilla supports
- We have "Remaining work per person" on iteration views, but it's empty for finished iterations, so it's hard to retrospectively get information on capacity for a given iteration. Would be good to add "Total work per person" showing finished bugs (including the total for the whole team) in addition to the one showing unfinished bugs
- On the Engineering Allocation view, we could consider adding a section at the bottom showing the same stuff. But instead of remaining/total work per person, per iteration, it would probably be per release, as the Engineering Allocation view is per release.

aminomancer:
Thanks for reviewing!

AllegroFox:
To your point about the XL: looks like the max number if points in Bugzilla is 15; which we could just use for XL regardless despite it not being in sequence (that would leave us 8 & 13 for L). Seems reasonable as a workaround for 20+ not being in bugzilla, since we would end up with 2 point amounts for each "level" (1 & 2 for SM, 3 & 5 for MD, 8 & 13 for L) and then just max value for XL.

I like the idea of "total work per person/per iteration" esp. since the pointing is meant to be a little retroactive, relative to previous amounts of work. We can give this change a little time to percolate and see what people think, and see if that would make the process a little clearer for folks

PR #294, opened 2024-02-23
New Metas view, add logo header, new styles, cleanup old cruft

### LINKED ISSUE #289: [FR] Stop tracking non-OMC bugs in Bugzy Triage
We're currently [tracking 4 components](https://github.com/mozilla/bugzy/blob/bfd5b67bb4c3af072d73699cf3e1dc7de2c90328/src/config/project_settings.js#L5-L10) in Triage, 3 of which we are not actually responsible for. No idea if the teams responsible for New Tab Page, Pocket, or Nimbus Desktop Client are using or will use Bugzy for triage, but if so, we should either separate our instance from theirs, or at least create separate views for separate teams. As it is, adding these 3 components is one of the main reasons the Triage view takes so long to load - it quadruples the size of the query.

Currently, the triage view is divided into 4 tabs according to these components. It should just be a single tab for now, meaning tabs should be hidden, at least until we add more components or something. It could be that we add some GitHub integration where, for example, Bugzy issues are tracked on a Bugzy tab in the triage view. Then we might benefit from showing tabs. But even that I'm not sure of. If we were to start triaging GitHub issues on bugzy, I think we'd probably want them in the same view as everything else, just to ensure we don't lose track of them.

aminomancer:
I tried disentangling these commits into separate PRs, but it's a bit too complicated, with too many overlapping changes. I think it's best to just merge without squashing, so that each commit is preserved individually in the history.

### Features

- Add a Metas view where you can find meta bugs that are not used commonly enough to be included in the sidebar. That way you can probably find a meta for your bug even if you're working on something obscure. This should also help us revive metas that are still relevant, and close metas that are no longer relevant.
- Show a warning color/underline and tooltip on the BMO links for bugs that are missing a meta bug. We have enough metas that all bugs should really have a meta, and it's easier to add them now that we have a Metas view.

### Style

- Add a logo header to the top of the sidebar
- Use Mozilla's fonts and typography guidelines (using Google WebFonts, which conveniently has all the Mozilla fonts)
- Fix the broken feature that was supposed to highlight bugs that are assigned to you, and make it stand out more by using a light blue background color instead of slightly bolder type
- Adjust the visual appearance of a few minor elements
- Use the proper ellipsis character instead of three periods
- Adjust priority colors slightly

### Fix/Refactor
- Switch to YAML config files where possible
- Change the Issues link on the About page to lead to the proper Issues page, not the "New Issue" page

### Cleanup

- Stop tracking non-OMC components (Fixes #289)
- Remove UI for UX/QA needinfos, as the email lists were way out of date and not feasible for us to maintain.
    - I asked around and received an API example for automatically fetching a list of UX designers and QA engineers from people.mozilla.org, based on a list of staff managed by a given manager. However, we can add this later if we think it's important.
- Remove some unused code
- Switch from `BugList::render` to `BugList::children` for a few components


nuggetsnegin:
I'll take a look through this tomorrow morning! 👍🏻

AllegroFox:
Re: removing the unused components - I remember that the last time I spoke to Scott about it (it was a while ago tbf) the Pocket folks were still actively using Bugzy for triage. I still think we should do this, but it would be nice to reach out and give them a heads up && ask if they're up to maintaining their own fork. I've pinged Scott to see if he knows :)

src/content/components/ActiveRSMessages/ActiveRSMessages.module.scss
```diff
@@ -1,17 +1,21 @@
 .RSTable {
-  padding: .6em;
+  padding: 0.6em;
   margin: 0 auto;
   width: max-content;
```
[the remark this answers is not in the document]
aminomancer: Filed #297

src/content/components/JiraView/JiraView.js
AllegroFox, line 1: JiraView looks broken right now; possibly because `this.state.JiraTickets` is undefined at the time of render

aminomancer: Nice catch, this one was missing default state for some reason. Fixed in 6e94751

AllegroFox (approved):
It's looking good! You addressed everything I noticed on the fly so this is just R+ :)

(we should probably have some tests tho)

aminomancer:
> It's looking good! You addressed everything I noticed on the fly so this is just R+ :)

Thanks!

> (we should probably have some tests tho)

Agreed - filed #298

PR #299, opened 2024-03-01
feat: restore resolved feature view

aminomancer:
There used to be a QA tab on feature views, which displayed resolved bugs. I recently removed it to speed up the bugzilla query for feature views. This PR restores it. I also added resolved date sorting, since previously the bugs in this view were unsorted.

### DIFF
--- a/src/content/components/FeatureView/FeatureView.js
+++ b/src/content/components/FeatureView/FeatureView.js
@@ -1,9 +1,11 @@
-import React from "react";
+import React, { useContext, useEffect, useMemo, useState } from "react";
 import { BugList } from "../BugList/BugList";
 import { CopyButton } from "../CopyButton/CopyButton";
+import { isBugResolved } from "../../lib/utils";
 import { GlobalContext } from "../GlobalContext/GlobalContext";
 import { Container } from "../ui/Container/Container";
-import { MiniLoader } from "../Loader/Loader";
+import { Tabs } from "../ui/Tabs/Tabs";
+import { Loader, MiniLoader } from "../Loader/Loader";
 import { removeMeta } from "../../../common/removeMeta";
 import * as gStyles from "../../styles/gStyles.module.scss";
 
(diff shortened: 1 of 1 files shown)


src/content/components/FeatureView/FeatureView.js
```diff
@@ -310,121 +426,193 @@ export class FeatureView extends React.PureComponent {
...
+        crossOutResolved={false}
+        bugs={bugs.releaseResolved}
+        columns={[...resolvedDisplayColumns, "target_milestone"]}
+      />
+      <FeatureBugList
+        title="Other"
+        hideIfEmpty={true}
+        crossOutResolved={false}
+        bugs={bugs.resolved}
+        columns={[...resolvedDisplayColumns, "target_milestone"]}
```
punamdahiya, line 505: For Other section we can use `cf_fx_iteration`  in place of `target_milestone` which stays `---` for Other section by changing Line 380 above to use `displayColumnFilter`

```
  const resolvedFieldFilter = f =>
    !["cf_fx_points"].includes(f);

  const displayColumnFilter = f =>
    !["cf_fx_points", "cf_fx_iteration"].includes(f);
  
const resolvedDisplayColumns = useMemo(
    () => displayColumns.filter(displayColumnFilter),
    []
  );
```

punamdahiya (approved):
Thanks for Quick turnaround! With feedback addressed looks good to merge.

src/content/components/FeatureView/FeatureView.js
```diff
@@ -310,121 +426,193 @@ export class FeatureView extends React.PureComponent {
...
+    [currentRelease, prevRelease, upliftTrackingField]
+  );
 
-    return (
-      <Container
-        loaded={this.state.loaded}
-        heading={
+  return (
+    <Container
+      loaded={true}
```
punamdahiya, line 571: Noticed Remaining/Open View doesn't refresh when switching between features and above loaded state needs to maintained to force refresh

src/content/components/FeatureView/FeatureView.js
```diff
@@ -310,121 +426,193 @@ export class FeatureView extends React.PureComponent {
       };
     };
 
-    const predicate = () =>
-      this._isMounted && id === this.props.match.params.id;
-    await this.context.qm.runCachedQueries(
+    const predicate = () => isMounted;
+    qm.runCachedQueries(
       subMetaQuery,
```
punamdahiya, line 431: Curious why we need `subMetaQuery` for ResolvedView?

aminomancer: The way the feature view works, both for ResolvedView and EngineeringView, is that it fetches all the blockers of the given feature bug and displays them. But some of those blockers are submetas. So it first fetches submeta blockers and uses those as the input to the next query, to get all blockers of submeta blockers of the feature bug.

EngineeringView additionally sorts the bugs by submeta, to create categories within the view. ResolvedView doesn't do any sorting by submeta, but it still needs to query the submetas since it's trying to show blockers of blockers.

I think a nicer way to do this would be some kind of recursive query, where it queries blockers of the feature meta, and if any of those blockers have the `meta` keyword, it adds them to a second query to get their blockers, and if any of _those_ blockers have the `meta` keyword, it does the same thing again; with perhaps up to 4 levels of recursion. Right now it only shows 2 levels, i.e. blockers of blockers of the feature.

But since I was just trying to restore this view ASAP so it'll be available for us, I left the query the same as it was before. Adding that kind of recursion will be more time-intensive, since I'd need to figure out visually how to represent up to 4 levels, or however many we choose to show. So probably something for a followup.

src/content/components/FeatureView/FeatureView.js
aminomancer: Good catch, I forgot to add `key={metaId}`

PR #303, opened 2024-03-04
Add arrow buttons on either side of the title on IterationView to go to previous/next iteration

### LINKED ISSUE #291: [FR] Add arrow buttons on either side of the title on IterationView to go to previous/next iteration
In [an iteration view](https://www.bugzy.org/iteration/123.3), the title displays the iteration name. On the left side of this title, there could be a button leading to the previous iteration (in this case 123.2) and to the right, a button leading to the next iteration (124.1). Just simple left/right arrows would do it.

aminomancer:
Fixes #291

### DIFF
--- a/src/content/components/IterationView/IterationView.tsx
+++ b/src/content/components/IterationView/IterationView.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { Link } from "react-router-dom";
 import { GlobalContext, MetaBug } from "../GlobalContext/GlobalContext";
 import { BugList } from "../BugList/BugList";
 import { useBugFetcher, Bug, BugQuery } from "../../hooks/useBugFetcher";
@@ -9,6 +10,7 @@ import { CompletionBar } from "../CompletionBar/CompletionBar";
 import { isBugResolved } from "../../lib/utils";
 import { teams as emailLists } from "../../../config/people";
 import { BUGZILLA_TRIAGE_COMPONENTS } from "../../../config/project_settings";
+import * as styles from "./IterationView.module.scss";
 import * as priorityStyles from "../PriorityGuide/PriorityGuide.module.scss";
 
 interface GetQueryOptions {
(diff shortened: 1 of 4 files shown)


punamdahiya:
Looks like some patches are committed directly to master which conflicts and this PR needs to be rebased with latest master

aminomancer:
Rebased, should apply cleanly now

src/content/components/IterationView/IterationView.tsx
```diff
@@ -227,8 +226,46 @@ export const IterationView: React.FunctionComponent<
     },
     [metas]
   );
+  const heading = React.useMemo(() => {
+    const nextIteration = iterations.getAdjacentIteration(1, props.iteration);
```
punamdahiya, line 230: For current iteration view, we can hide next arrow as it's mostly empty and if really needed can be reached from select iteration drop down.
```

 const nextIteration =
      props.iteration !== props.currentIteration.number
        ? iterations.getAdjacentIteration(1, props.iteration)
        : null;
```

src/content/components/Router/Router.js
```diff
@@ -508,6 +508,11 @@ export class Router extends React.PureComponent {
               <Route exact={true} path="/">
```
punamdahiya, line 508: With easily accessible arrows and Select iteration drop down, we can remove `Next Iteration` menu item in left nav (Line 163 above)

punamdahiya (approved):
Looks great! Left feedback on further optimizing next iteration flow. Thanks for the patch!

src/content/components/Router/Router.js
aminomancer: Good idea. But if we remove the next arrow from the current iteration view, then there won't be a way to reach Next Iteration except from the dropdown picker. So do you still want to do both? Remove the next arrow from current iteration _and_ remove `Next Iteration` menu item? Might make sense to just remove one or the other

punamdahiya: IMO `Next Iteration` view has usually less to no bugs and having access from select dropdown should suffice.

aminomancer:
Great, thanks! Resolved feedback.

PR #316, opened 2024-05-07
Issue 266 - Include "No Bugs" placeholder when there are no bugs in General Triage

### LINKED ISSUE #266: Add a placeholder message to display when there are no bugs in a view
In the case of the General Triage (#264) view, we hide the buglists when they're empty, since there are 6 buglists and they often are individually empty. But right now, the entire view is empty, as all 6 queries return no bugs. So we get a blank page, which might look like a bug without a placeholder saying "no bugs" or something.

sarahhjchung:
Fixes #266 

Changes made: 
- Included a "No bugs found." placeholder when all the bug lists are empty in the General Triage view

Screenshot 2024-05-08 at 1 44 59 PM


aminomancer (approved):
Thanks!

aminomancer (changes requested):
Code looks good and is working, but after running it myself, I think the header needs some work visually.

- It's left aligned
- It's not very big
- The string should be `No bugs found.` rather than `No Bugs` (see BugList.js)

Check `.emptyState` in the BugList and FeatureView stylesheets, that would be good for consistency (add a .module.scss stylesheet).

A visual mockup -

image

aminomancer (approved):
Thanks!

PR #321, opened 2024-06-19
Allow viewing meta bugs as bugzy features even if they don't block the bugzy epic

### LINKED ISSUE #320: [FR] Allow viewing meta bugs as bugzy features even if they don't block the bugzy epic
This will be a nice QoL feature. We show a limited number of high priority meta bugs in the sidebar, but there are many more metas not included there. We added a Metas view for them in #294, but the links still only go to Bugzilla. I want those links to lead to FeatureViews instead, since it doesn't cost us anything to allow querying and rendering them in Bugzy.

They still shouldn't show up the sidebar, but if you navigate to them via the Metas view, you'll stay within Bugzy rather than navigating to Bugzilla. Then you can navigate to the Bugzilla bug with the header link. Or make it so the bug number links in the Metas view lead to Bugzilla, but the summary links lead to Bugzy FeatureViews. That seems like the most elegant approach.

aminomancer:
Fixes #320

### DIFF
--- a/src/content/components/Router/Router.js
+++ b/src/content/components/Router/Router.js
@@ -3,6 +3,7 @@ import * as styles from "./Router.module.scss";
 import { BugListView } from "../BugListView/BugListView";
 import {
   BrowserRouter,
+  Link,
   NavLink,
   Redirect,
   Route,
@@ -116,9 +117,7 @@ export class Router extends React.PureComponent {
       })
       .map(meta => ({
         path: `/feature/${meta.id}`,
-        label: `${meta.priority !== "P1" ? `[${meta.priority}] ` : ""}${
-          meta.displayName
-        }`,
+        label: meta.displayName,
       }));
   }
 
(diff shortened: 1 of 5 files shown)


src/content/components/Router/Router.js
```diff
@@ -116,9 +117,7 @@ export class Router extends React.PureComponent {
       })
       .map(meta => ({
         path: `/feature/${meta.id}`,
-        label: `${meta.priority !== "P1" ? `[${meta.priority}] ` : ""}${
```
aminomancer, line 119: This wasn't necessary because the filter above at line 102 makes it impossible for the priority to not be P1. We only show unresolved P1 bugs that block the bugzy epic.

nuggetsnegin (approved):
LGTM and works as expected locally (and sorry for missing this PR!)! This will be such a nice QOL change. Not really in the scope of this PR so I can file another issue but what do you think of making the Metas a button somewhere... (maybe underneath All Features as `All Metas`) in:

immagine

I find that it's a bit hidden as a link and I actually never routed to it before and realized those metas linked to bugzilla.. 😅 But maybe this is an issue on my end - I just wonder if it needs to be nested under `In-product Messaging – Firefox x...`

aminomancer:
Thanks! Moving it up to the main links would be reasonable, now that we've removed Pocket and Nimbus. But I haven't spun up the bugzy repo since I made this PR and just don't have much time to get back into it.

